PR #56, opened 2022-08-24
config: dynamic proxy config

xhebox:

Issue Number: ref #57

Problem Summary: Dynamic config `keep alive` and `max connections` for proxy server.

1. In short, added background service for config manager to poll online config regularlly/watch etcd event. It will then be applied to the running server.
2. API return value minor improvement. Added `/api/admin/config/proxy` for online proxy config.
3. Configs have `yaml, toml, json` all struct tags
4. `errors.Wrap(f)` will return nil for nil error
5. Makefile minor improvement

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
./bin/weirproxy ...
echo '{"max_connections": 0}' | ./bin/weirctl config proxy
zero connection error
echo '{"max_connections": 1}' | ./bin/weirctl config proxy

start two mysql client, and one of them will be rejected by the server
```
- [ ] No code

Notable changes

- [x] Has weirctl change

### Release note

```release-note
Support dynamic configuration for some proxy server options
```

### DIFF
--- a/pkg/config/proxy.go
+++ b/pkg/config/proxy.go
@@ -52,7 +52,8 @@ type Metrics struct {
 }
 
 type ConfigManager struct {
-	IgnoreWrongNamespace bool `yaml:"ignore_wrong_namespace" toml:"ignore_wrong_namespace" json:"ignore_wrong_namespace"`
+	IgnoreWrongNamespace bool   `yaml:"ignore_wrong_namespace" toml:"ignore_wrong_namespace" json:"ignore_wrong_namespace"`
+	WatchInterval        string `yaml:"watch_interval" toml:"watch_interval" json:"watch_interval"`
 }
 
 type ProxyServerOnline struct {
--- a/pkg/config/proxy.go
+++ b/pkg/config/proxy.go
@@ -28,64 +28,68 @@ const (
 )
 
 type Config struct {
-	Workdir string `yaml:"workdir"`
+	Workdir string `yaml:"workdir" toml:"workdir" json:"workdir"`
 
-	LCUrlsI []string  `yaml:"listen-urls"`
-	ACUrlsI []string  `yaml:"advertise-urls"`
-	LPUrlsI []string  `yaml:"listen-peer-urls"`
-	APUrlsI []string  `yaml:"advertise-peer-urls"`
-	LCUrls  []url.URL `yaml:"-"`
-	ACUrls  []url.URL `yaml:"-"`
-	LPUrls  []url.URL `yaml:"-"`
-	APUrls  []url.URL `yaml:"-"`
+	LCUrlsI []string  `yaml:"listen-urls" toml:"listen-urls" json:"listen-urls"`
+	ACUrlsI []string  `yaml:"advertise-urls" toml:"advertise-urls" json:"advertise-urls"`
+	LPUrlsI []string  `yaml:"listen-peer-urls" toml:"listen-peer-urls" json:"listen-peer-urls"`
+	APUrlsI []string  `yaml:"advertise-peer-urls" toml:"advertise-peer-urls" json:"advertise-peer-urls"`
+	LCUrls  []url.URL `yaml:"-" toml:"-" json:"-"`
+	ACUrls  []url.URL `yaml:"-" toml:"-" json:"-"`
+	LPUrls  []url.URL `yaml:"-" toml:"-" json:"-"`
+	APUrls  []url.URL `yaml:"-" toml:"-" json:"-"`
 
-	Config   ConfigManager `yaml:"config"`
-	Proxy    ProxyServer   `yaml:"proxy"`
-	API      API           `yaml:"api"`
-	Metrics  Metrics       `yaml:"metrics"`
-	Log      Log           `yaml:"log"`
-	Security Security      `yaml:"security"`
+	Config   ConfigManager `yaml:"config" toml:"config" json:"config"`
+	Proxy    ProxyServer   `yaml:"proxy" toml:"proxy" json:"proxy"`
+	API      API           `yaml:"api" toml:"api" json:"api"`
+	Metrics  Metrics       `yaml:"metrics" toml:"metrics" json:"metrics"`
+	Log      Log           `yaml:"log" toml:"log" json:"log"`
+	Security Security      `yaml:"security" toml:"security" json:"security"`
 }
 
 type Metrics struct {
-	PromCluster string `yaml:"prom_cluster"`
+	PromCluster string `yaml:"prom_cluster" toml:"prom_cluster" json:"prom_cluster"`
 }
 
 type ConfigManager struct {
-	IgnoreWrongNamespace bool `yaml:"ignore_wrong_namespace"`
+	IgnoreWrongNamespace bool `yaml:"ignore_wrong_namespace" toml:"ignore_wrong_namespace" json:"ignore_wrong_namespace"`
+}
+
+type ProxyServerOnline struct {
+	MaxConnections uint64 `yaml:"max_connections" toml:"max_connections" json:"max_connections"`
+	TCPKeepAlive   bool   `yaml:"tcp_keep_alive" toml:"tcp_keep_alive" json:"tcp_keep_alive"`
 }
 
 type ProxyServer struct {
-	Addr           string `yaml:"addr"`
-	MaxConnections uint64 `yaml:"max_connections"`
-	TCPKeepAlive   bool   `yaml:"tcp_keep_alive"`
-	PDAddrs        string `yaml:"pd_addrs"`
-	ProxyProtocol  string `yaml:"proxy_protocol"`
+	ProxyServerOnline
+	Addr          string `yaml:"addr" toml:"addr" json:"addr"`
+	PDAddrs       string `yaml:"pd_addrs" toml:"pd_addrs" json:"pd_addrs"`
+	ProxyProtocol string `yaml:"proxy_protocol" toml:"proxy_protocol" json:"proxy_protocol"`
 }
 
 type API struct {
-	EnableBasicAuth bool   `yaml:"enable_basic_auth"`
-	User            string `yaml:"user"`
-	Password        string `yaml:"password"`
+	EnableBasicAuth bool   `yaml:"enable_basic_auth" toml:"enable_basic_auth" json:"enable_basic_auth"`
+	User            string `yaml:"user" toml:"user" json:"user"`
+	Password        string `yaml:"password" toml:"password" json:"password"`
 }
 
 type Log struct {
-	Level   string  `yaml:"level"`
-	Encoder string  `yaml:"encoder"`
-	LogFile LogFile `yaml:"log_file"`
+	Level   string  `yaml:"level" toml:"level" json:"level"`
+	Encoder string  `yaml:"encoder" toml:"encoder" json:"encoder"`
+	LogFile LogFile `yaml:"log_file" toml:"log_file" json:"log_file"`
 }
 
 type LogFile struct {
-	Filename   string `yaml:"filename"`
-	MaxSize    int    `yaml:"max_size"`
-	MaxDays    int    `yaml:"max_days"`
-	MaxBackups int    `yaml:"max_backups"`
+	Filename   string `yaml:"filename" toml:"filename" json:"filename"`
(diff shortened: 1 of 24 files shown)


djshow832 (commented):
https://github.com/pingcap/TiProxy/blob/a9b15514f5f0bc282b35678df69b0ab07f053b61/cmd/weirctl/util.go#L46
I think it should be `for _, i := range rand.Perm(len(bctx.CUrls)) {`?

xhebox:
> https://github.com/pingcap/TiProxy/blob/a9b15514f5f0bc282b35678df69b0ab07f053b61/cmd/weirctl/util.go#L46
> 
> I think it should be `for _, i := range rand.Perm(len(bctx.CUrls)) {`?

Nope, I am using it as `bctx.CUrls[i]`. If you like this style, I can change it.

pkg/manager/config/proxy.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+
+	"github.com/pingcap/TiProxy/pkg/config"
+	"github.com/pingcap/TiProxy/pkg/util/errors"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+)
+
+var (
+	ErrZeroMaxConn = errors.New("zero max connection set")
```
djshow832, line 28: Typically, max_connections=0 means unlimited.

pkg/manager/config/manager.go
```diff
@@ -106,16 +158,7 @@ func (e *ConfigManager) del(ctx context.Context, ns, key string) error {
...
-}
-
-func appendSlashToDirPath(dir string) string {
-	if len(dir) == 0 || dir[len(dir)-1] == '/' {
-		return dir
-	}
-	return dir + "/"
+	e.cancel()
+	e.wg.Wait()
+	return errors.Wrapf(e.etcdClient.Close(), "fail to close config manager")
```
djshow832, line 163: e.etcdClient is never closed

pkg/manager/config/manager.go
```diff
@@ -59,21 +71,61 @@ func (srv *ConfigManager) Init(addrs []string, cfg config.ConfigManager, logger
...
+	logger := e.logger.With(zap.String("comp", wkey))
+	e.wg.Run(func() {
+		wch := e.etcdClient.Watch(ctx, wkey)
+		ticker := time.Tick(DefaultWatchLease)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker:
+				resp, err := e.kv.Get(ctx, path.Join(e.basePath, ns, wkey))
```
djshow832, line 98: wkey=path.Join(e.basePath, ns, key), you are join e.basePath and ns again?
```suggestion
				resp, err := e.kv.Get(ctx, wkey)
```

pkg/manager/config/manager.go
```diff
@@ -59,21 +71,61 @@ func (srv *ConfigManager) Init(addrs []string, cfg config.ConfigManager, logger
...
+				if err != nil {
+					logger.Warn("failed to poll", zap.Error(err))
+					break
+				}
+				if len(resp.Kvs) != 1 {
+					logger.Warn("failed to poll", zap.Error(ErrNoOrMultiResults))
+					break
+				}
+			case res := <-wch:
+				if res.Canceled {
```
djshow832, line 108: When will it be canceled?

pkg/manager/config/manager.go
```diff
@@ -59,21 +71,61 @@ func (srv *ConfigManager) Init(addrs []string, cfg config.ConfigManager, logger
...
+	srv.cancel = cancel
+
+	srv.initProxyConfig(ctx)
+
 	return nil
 }
 
+func (e *ConfigManager) watch(ctx context.Context, ns, key string, f func(*zap.Logger, *clientv3.Event)) {
+	wkey := path.Join(e.basePath, ns, key)
+	logger := e.logger.With(zap.String("comp", wkey))
```
djshow832, line 89: What does `comp` mean?

pkg/manager/config/manager.go
```diff
@@ -59,21 +71,61 @@ func (srv *ConfigManager) Init(addrs []string, cfg config.ConfigManager, logger
...
+			case res := <-wch:
+				if res.Canceled {
+					logger.Warn("failed to watch", zap.Error(res.Err()))
+					wch = e.etcdClient.Watch(ctx, wkey)
+					break
+				}
+
+				for _, evt := range res.Events {
+					f(logger, evt)
+				}
```
djshow832, line 116: What about resetting the ticker after an event?

pkg/manager/config/manager.go
```diff
@@ -59,21 +71,61 @@ func (srv *ConfigManager) Init(addrs []string, cfg config.ConfigManager, logger
...
+
 	return nil
 }
 
+func (e *ConfigManager) watch(ctx context.Context, ns, key string, f func(*zap.Logger, *clientv3.Event)) {
+	wkey := path.Join(e.basePath, ns, key)
+	logger := e.logger.With(zap.String("comp", wkey))
+	e.wg.Run(func() {
+		wch := e.etcdClient.Watch(ctx, wkey)
+		ticker := time.Tick(DefaultWatchLease)
```
djshow832, line 92: I don't think it's a lease, it's just an interval.

djshow832:
> > https://github.com/pingcap/TiProxy/blob/a9b15514f5f0bc282b35678df69b0ab07f053b61/cmd/weirctl/util.go#L46
> > 
> > I think it should be `for _, i := range rand.Perm(len(bctx.CUrls)) {`?
> 
> Nope, I am using it as `bctx.CUrls[i]`. If you like this style, I can change it.

```go
for i := range rand.Perm(len(bctx.CUrls)) {
	req.URL.Host = bctx.CUrls[i]
```
is just equivalent to
```go
for _, host := bctx.CUrls {
	req.URL.Host = host
```
because `i` is always an increasing integer. It's the index of the array `rand.Perm(len(bctx.CUrls))`.

pkg/manager/config/manager.go
xhebox: When there is an error on etcd.

pkg/manager/config/manager.go
xhebox: Why?

pkg/manager/config/manager.go
xhebox: component, addressed.

xhebox:
> because `i` is always an increasing integer. It's the index of the array `rand.Perm(len(bctx.CUrls))`.

I know, I prefer using `i` style instead of `_, host`, changed.

xhebox:
> > > https://github.com/pingcap/TiProxy/blob/a9b15514f5f0bc282b35678df69b0ab07f053b61/cmd/weirctl/util.go#L46
> > > 
> > > I think it should be `for _, i := range rand.Perm(len(bctx.CUrls)) {`?
> > 
> > 
> > Nope, I am using it as `bctx.CUrls[i]`. If you like this style, I can change it.
> 
> ```go
> for i := range rand.Perm(len(bctx.CUrls)) {
> 	req.URL.Host = bctx.CUrls[i]
> ```
> 
> is just equivalent to
> 
> ```go
> for _, host := bctx.CUrls {
> 	req.URL.Host = host
> ```
> 
> because `i` is always an increasing integer. It's the index of the array `rand.Perm(len(bctx.CUrls))`.

Addressed.

pkg/manager/config/manager.go
djshow832: Because it needs to release resources. See what it does:
```go
func (c *Client) Close() error {
	c.cancel()
	if c.Watcher != nil {
		c.Watcher.Close()
	}
	if c.Lease != nil {
		c.Lease.Close()
	}
	if c.conn != nil {
		return toErr(c.ctx, c.conn.Close())
	}
	return c.ctx.Err()
}
```

xhebox: But there is `errors.Wrapf(e.etcdClient.Close(), ...)`?

pkg/manager/config/manager.go
xhebox: I think it is only testable by failpoints. To trigger the ticker manually.

pkg/manager/config/manager.go
djshow832: Oops, I missed it.

pkg/manager/config/manager.go
djshow832: Yes, it's hard to test. Or you can pass the interval as a parameter and set the interval to 1 second in the test.

xhebox: Added a test for `e.watch()`.

pkg/config/proxy.go
```diff
@@ -28,64 +28,69 @@ const (
...
 
 type Metrics struct {
-	PromCluster string `yaml:"prom_cluster"`
+	PromCluster string `yaml:"prom_cluster" toml:"prom_cluster" json:"prom_cluster"`
 }
 
 type ConfigManager struct {
-	IgnoreWrongNamespace bool `yaml:"ignore_wrong_namespace"`
+	IgnoreWrongNamespace bool   `yaml:"ignore_wrong_namespace" toml:"ignore_wrong_namespace" json:"ignore_wrong_namespace"`
+	WatchInterval        string `yaml:"watch_interval" toml:"watch_interval" json:"watch_interval"`
```
djshow832, line 56: I thought about this, it's hard to explain what this really means to the user.
For common cases, the config will be refreshed once updated. This interval is only used for defensive programming.
As for TiDB, the refresh interval for privileges and global variables are not exposed to users either, and nobody complains about it.

pkg/manager/config/manager.go
xhebox: > is it possible to be in a dead loop if etcd always fails

This is expected. It should be in the loop, until context is canceled. But yes, it should not consume much CPU usage.

The only case should be network rejection, where server rejected connections from the client actively. Common firewall will only drop packets. So it is a very rare case.

But anyway, sleeping for seconds or 1 min does not hurt.

pkg/config/proxy.go
xhebox: We could categorize them into **KNOW WHAT YOU ARE DOING options** and provide a resonable default. If one don't understand, just don't set.

You can choose to use the default good config, or make your decision. That is users' freedom. BTW, these are possibly useful for debugging online.

xhebox: > As for TiDB, the refresh interval for privileges and global variables are not exposed to users either, and nobody complains about it.

There is no choice. If we have a documentation said the priv interval is XXX. Maybe someone will complain that it is too short or too long.

pkg/proxy/proxy.go
```diff
@@ -113,7 +118,7 @@ func (s *SQLServer) onConn(ctx context.Context, conn net.Conn) {
 	tcpKeepAlive := s.mu.tcpKeepAlive
 
 	// 'maxConns == 0' => unlimited connections
-	if maxConns != 0 && maxConns <= conns {
+	if maxConns != 0 && maxConns < conns {
```
djshow832, line 121: It should be `maxConns <= conns` because `s.mu.clients[connID] = clientConn` is not executed yet.

pkg/config/proxy.go
djshow832: > > As for TiDB, the refresh interval for privileges and global variables are not exposed to users either, and nobody complains about it.
> 
> There is no choice. They don't even know about that. Maybe they have a guess in their mind, and thought it is correct. If we have a documentation said the priv interval is XXX. Maybe someone will complain that it is too short or too long.

If we make it as a config, we must document it, and then users are confused and complain. I sometimes encounter the situation where a user saw a config and doesn't understand what it means and then asks us. It just brings us trouble. If we're removing it, it again brings compatibility problems.

xhebox: I would not consider it a problem. First of all, TiProxy will have much less options. It is much simpler and not that hard to understand.

My point is that documentation of complex options can be **highlighted**, in **another section** or **another page** with a **large NOTE**. The example for these options can be separated into **another file**, or **commented by default**.

If users feel normal options are enough, they surely won't dig deep into these complicated options. Moreover, we recommend you not to use these options unless you understand it. TiDB docs has none of these guidance. And description for system variables usually do not have any external link to the related page or doc.

djshow832 (approved):

PR #61, opened 2022-08-30
util: try to fix panic

xhebox:

Issue Number: close None

Problem Summary: I don't know how it is produced. `len(xx) >= 0` is expected, thus the judgement should be true forever. Anyway, this is a try.

1. `EncodeEntry` will not free the buffer anymore, it is left to zap logger, based on https://github.com/uber-go/zap/blob/4a895a245ab794e3ae965783843a9fbc793083bc/zapcore/core.go#L95-L100
2. `StackTrace` by zap is multiline, this is not compatible with `tidb` format.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has weirctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/util/cmd/encoder.go
+++ b/pkg/util/cmd/encoder.go
@@ -50,7 +50,7 @@ func (c tidbEncoder) clone() *tidbEncoder {
 }
 
 func (c tidbEncoder) Clone() zapcore.Encoder {
-	return c.Clone()
+	return c.clone()
 }
 
 func (c *tidbEncoder) beginQuoteFiled() {
--- a/pkg/util/cmd/encoder.go
+++ b/pkg/util/cmd/encoder.go
@@ -329,10 +329,11 @@ func (s *tidbEncoder) closeOpenNamespaces() {
 
 /* array encoder part */
 func (s *tidbEncoder) addElementSeparator() {
-	if s.line.Len() <= 0 {
+	length := s.line.Len()
+	if length == 0 {
 		return
 	}
-	switch s.line.Bytes()[s.line.Len()-1] {
+	switch s.line.Bytes()[length-1] {
 	case '{', '[', ':', ',', ' ', '=':
 	default:
 		s.line.AppendByte(',')
--- a/pkg/util/cmd/encoder.go
+++ b/pkg/util/cmd/encoder.go
@@ -45,8 +45,12 @@ func NewTiDBEncoder(cfg zapcore.EncoderConfig) zapcore.Encoder {
 	return &tidbEncoder{cfg, 0, _pool.Get()}
 }
 
-func (c tidbEncoder) Clone() zapcore.Encoder {
+func (c tidbEncoder) clone() *tidbEncoder {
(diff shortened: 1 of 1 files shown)


pkg/util/cmd/encoder.go
```diff
@@ -327,11 +329,10 @@ func (s *tidbEncoder) closeOpenNamespaces() {
 
 /* array encoder part */
 func (s *tidbEncoder) addElementSeparator() {
-	length := s.line.Len()
-	if length == 0 {
+	if s.line.Len() <= 0 {
 		return
 	}
-	switch s.line.Bytes()[length-1] {
+	switch s.line.Bytes()[s.line.Len()-1] {
```
djshow832, line 335: I don't think this makes any difference. If it's not thread-safe, it will still panic.

xhebox: Encoder is not thread-safe, same for pingcap/log or zap. I'll guess the root cause is `EncodeEntry`. The modifications here may not make any difference. I just change it for changes.

xhebox: It is reverted.

pkg/util/cmd/encoder.go
```diff
@@ -74,10 +78,8 @@ func (c *tidbEncoder) encodeError(f zapcore.Field) {
 		}
 	}
 }
-func (c *tidbEncoder) EncodeEntry(ent zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
-	c.line.Free()
-	c.line = _pool.Get()
-
+func (e *tidbEncoder) EncodeEntry(ent zapcore.Entry, fields []zapcore.Field) (*buffer.Buffer, error) {
+	c := e.clone()
```
djshow832, line 82: In
https://github.com/uber-go/zap/blob/4a895a245ab794e3ae965783843a9fbc793083bc/zapcore/json_encoder.go#L365
and
https://github.com/pingcap/log/blob/master/zap_text_encoder.go#L406
they are both `e.clone()`. So you modified it?

pkg/util/cmd/encoder.go
```diff
@@ -45,10 +45,14 @@ func NewTiDBEncoder(cfg zapcore.EncoderConfig) zapcore.Encoder {
 	return &tidbEncoder{cfg, 0, _pool.Get()}
 }
 
-func (c tidbEncoder) Clone() zapcore.Encoder {
+func (c tidbEncoder) clone() *tidbEncoder {
 	return &tidbEncoder{c.EncoderConfig, 0, _pool.Get()}
 }
 
+func (c tidbEncoder) Clone() zapcore.Encoder {
+	return c.Clone()
```
djshow832, line 53:
```suggestion
	return c.clone()
```
Are you sure it calls itself?

xhebox: I should have changed this to `c.clone()`. Maybe I made mistakes in `--rebase`.

djshow832 (approved):

PR #65, opened 2022-08-31
namespace: create default namespace at the first run

### LINKED ISSUE #47: Create a default namespace if the namespace configuration is not specified
Currently, users must specify `--namespaces conf/namespace` to start the proxy for the first time, otherwise users cannot connect to it.
For testing, users can just omit the namespace configuration. The proxy can automatically create a default one for simplification.
This should also apply to `--config=conf/weirproxy.yaml`.

xhebox:

Issue Number: close #47

Problem Summary: As title.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has weirctl change

### Release note

```release-note
None
```


pkg/server/server.go
```diff
@@ -129,6 +132,20 @@ func NewServer(ctx context.Context, cfg *config.Config, logger *zap.Logger, name
 			return
 		}
 
+		if errors.Is(dirErr, os.ErrNotExist) {
```
djshow832, line 135: Maybe it's a better idea to check whether there are any namespaces after etcd starts? Is it easy to implement?
`work` exists doesn't really mean the etcd database exists.

xhebox: Yes, we can implement it with an extra RPC.

pkg/server/server.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
```
djshow832, line 21: Please compile before committing.

pkg/server/server.go
```diff
@@ -129,6 +129,25 @@ func NewServer(ctx context.Context, cfg *config.Config, logger *zap.Logger, name
 			return
 		}
 
+		nscs, nerr := srv.ConfigManager.ListAllNamespace(ctx)
```
djshow832, line 132: Why add a default namespace before `ImportNamespaceFromDir`? What if the user specifies a directory?

xhebox: I'll remove `--namespaces` from server, but add it to the cli later.

djshow832: Why? There is no way to specify an initial namespace. Maybe the user doesn't want a default one.

xhebox: > Why? There is no way to specify an initial namespace. 

This is "good default preset". And if you need something different, use ctl to config it. I don't want to be implicit that `--namespaces` will override the default fallback.

> Maybe the user doesn't want a default one.

We can delete the default namespace using `weirctl` whenever we want. If we want to delete it permanently, then we must only init the fallback only once, i.e. detect the workdir.

xhebox: It is now changed to:

1. give `default` fallback on the first run by detecting `cfg.Workdir`
2. `./weirctl namespace import conf/namespace` if yout need to init something different
3. `./weirctl namespace del default` if you don't want the fallback
4. `""` is renamed to `"default"`. And empty namespace name is forbidden

djshow832: Is `weirctl` available in the cloud?

xhebox: > Is `weirctl` available in the cloud?

Honestly, I don't know. I haven't used DBAAS. But I think `pd ctl` is avaiable, so it should be avaiable.

pkg/server/server.go
```diff
@@ -129,9 +132,16 @@ func NewServer(ctx context.Context, cfg *config.Config, logger *zap.Logger, name
...
 
-		if namespaceFiles != "" {
-			err = srv.ConfigManager.ImportNamespaceFromDir(ctx, namespaceFiles)
-			if err != nil {
+		if errors.Is(dirErr, os.ErrNotExist) {
+			// first time running
+			nsc := &config.Namespace{
+				Namespace: "default",
+				Backend: config.BackendNamespace{
+					Instances:    []string{"127.0.0.1:4000"},
```
djshow832, line 140: My intention is to easily test without specifying a namespace directory.
If you insist to always create a default namespace (like a system namespace), at least don't use configurations like this. This seems like used for testing.

xhebox: Do you have any suggestion? PD and security is all configured in the config file. Backend only makes sense without setting pd. I can not come up with any better default configuration.

djshow832: You can leave it empty. Backend instances are only used for testing when there is no PD.

xhebox: Done.

conf/namespace/test_namespace.yaml
```diff
@@ -1,12 +1,8 @@
 namespace: "test_namespace"
 frontend:
   security:
-    cert: "./ttt"
```
djshow832, line 4: Now that we have a `default.yaml`, I don't think we need `test_namespace.yaml` anymore.

pkg/server/server.go
djshow832: Why change it back to check the dir?

xhebox: Because you said, one may don't want the default fallback. IMO, that means the proxy should fail if no namespace matched. But the default fallback works by detecting if there is any namespace. So there is always a default fallback.

I've changed it back to ensure it is only ran once. But after a second thought, I think we can solve this problem by support PD per namespace later. An empty namespace will prevent client connections, and only a non-empty/effective fallback will serve as the real "fallback".

PR #90, opened 2022-09-21
proxy/client: do not print error on EOF

xhebox:

Issue Number: close None

Problem Summary: Another logging improvement. `EOF` is normal and should be seen as error.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has weirctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/proxy/client/client_conn.go
+++ b/pkg/proxy/client/client_conn.go
@@ -28,6 +28,10 @@ import (
 	"go.uber.org/zap"
 )
 
+var (
+	ErrClientConn = errors.New("this is an error from client")
+)
+
 type ClientConnection struct {
 	logger            *zap.Logger
 	frontendTLSConfig *tls.Config    // the TLS config to connect to clients.
@@ -39,7 +43,7 @@ type ClientConnection struct {
 }
 
 func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager) *ClientConnection {
-	pkt := pnet.NewPacketIO(conn)
+	pkt := pnet.NewPacketIOWrapErr(conn, ErrClientConn)
 	return &ClientConnection{
 		logger:            logger,
 		frontendTLSConfig: frontendTLSConfig,
@@ -77,7 +81,9 @@ func (cc *ClientConnection) Run(ctx context.Context) {
 		return
 	}
 
-	if err := cc.processMsg(ctx); err != nil && !errors.Is(err, io.EOF) {
+	if err := cc.processMsg(ctx); err != nil &&
+		// ignore EOF from client
+		!(errors.Is(err, ErrClientConn) && errors.Is(err, io.EOF)) {
 		cc.logger.Info("process message fails", zap.String("remoteAddr", cc.Addr()), zap.Error(err))
 	}
 }
--- a/pkg/proxy/client/client_conn.go
+++ b/pkg/proxy/client/client_conn.go
@@ -17,6 +17,7 @@ package client
 import (
 	"context"
 	"crypto/tls"
+	"io"
 	"net"
 
 	"github.com/pingcap/TiProxy/lib/util/errors"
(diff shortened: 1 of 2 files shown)


pkg/proxy/client/client_conn.go
```diff
@@ -76,7 +77,7 @@ func (cc *ClientConnection) Run(ctx context.Context) {
 		return
 	}
 
-	if err := cc.processMsg(ctx); err != nil {
+	if err := cc.processMsg(ctx); err != nil && !errors.Is(err, io.EOF) {
```
djshow832, line 80: I need to know whether it's disconnected by the server or the client.
In many cases I encountered, the user reports that the client disconnects with the server and suspects it's the problem of the server. I need to check the log to confirm that it's the problem of the client because the client quits first.

xhebox: Got. I modified the code a little bit to wrap errors from `client` with `ErrClientConn`.

So if `EOF` comes from `client`, it is ignored. If it comes from `server`, it is logged.

djshow832: I think it's better to log even if `EOF` comes from the client. I can directly judge from one log file instead of searching all log files to confirm that there's no error from the server.

pkg/proxy/client/client_conn.go
```diff
@@ -76,7 +81,9 @@ func (cc *ClientConnection) Run(ctx context.Context) {
 		return
 	}
 
-	if err := cc.processMsg(ctx); err != nil {
+	if err := cc.processMsg(ctx); err != nil &&
+		// ignore EOF from client
+		!(errors.Is(err, ErrClientConn) && errors.Is(err, io.EOF)) {
```
djshow832, line 86: There may occur various disconnection errors. Maybe using `IsDisconnectError` is better.

pkg/proxy/client/client_conn.go
xhebox: >  I can directly judge from one log file instead of searching all log files to confirm that there's no error from the server.

That argument does not hold, if need to know if there is error from server. You need to check that all errors are from clients, in another words. You can just check if this connection is from server or client.

Whatever, I added `clientErr=boolean` and `serverErr=boolean` further. You can now `grep -r -H serverErr=true` to check that, without the need of logging `EOF` from the client.

pkg/proxy/client/client_conn.go
xhebox: Done.

pkg/proxy/client/client_conn.go
```diff
@@ -77,7 +81,10 @@ func (cc *ClientConnection) Run(ctx context.Context) {
 	}
 
 	if err := cc.processMsg(ctx); err != nil {
-		cc.logger.Info("process message fails", zap.String("remoteAddr", cc.Addr()), zap.Error(err))
+		clientErr := errors.Is(err, ErrClientConn)
+		if !(clientErr && pnet.IsDisconnectError(err)) {
```
djshow832, line 85: If `clientErr && pnet.IsDisconnectError(err) == true`, it's still not printed.

xhebox: So why not ignore `client disconnect` errors?  Even if you print it out, you still need to search all files to conclude that there is no `server disconnect` error by realizing all errors are from `client`.

I'll give another modification, `EOF` is a normal disconnection state that should be ignored. And `broken pipe` or `connection reset` will be logged.

xhebox: BTW, there is another argument. Since server will nerver end the connection in the command phase, thus disconnection must come from clients or network errors. Now all network errors are logged. If no errors are logged, then it is  disconnected by`EOF` from clients obviously.

djshow832: I don't think `EOF` is normal. Sending a `STMT_QUIT` is normal.

djshow832: It came up to me that there's always another disconnection log. So ignoring the `EOF` may be fine.

djshow832 (approved):

PR #91, opened 2022-09-21
*: fix hang bug

xhebox:

Issue Number: close None

Problem Summary: Open a client conenction. Close `tiproxy` without interrupting the connection, tiproxy will hang.

1. `MError` should not have `Unwrap` method. It is unsuitable to unwrap multiple errors into one specific error. But unwrapping to itself may cause infinite recursion, because some functions will unwrap forever if he can.
2. While `router` and `backendObserver` does not inherit the cancel context, we can close `namespaces`, which will try to close routers.
3. No need to `ProxyServer.wg.Wait()`. We already waited at `Close()`.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
./bin/tiproxy
mycli -u root -P 6000 // in another terminal
type ctrl+c for tiproxy and see it hangs or not
```
- [ ] No code

Notable changes

- [ ] Has weirctl change

### Release note

```release-note
None
```

### DIFF
--- a/lib/util/errors/merror.go
+++ b/lib/util/errors/merror.go
@@ -61,12 +61,12 @@ func (e *MError) Error() string {
 	return fmt.Sprintf("%s", e)
 }
 
-func (e *MError) Unwrap() error {
-	return e
-}
-
 func (e *MError) Is(s error) bool {
-	return errors.Is(e.cerr, s)
+	is := errors.Is(e.cerr, s)
+	for _, e := range e.uerr {
+		is  = is || errors.Is(e, s)
+	}
+	return is
 }
 
 func (e *MError) Cause() []error {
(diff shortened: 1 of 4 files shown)


lib/util/errors/merror.go
```diff
@@ -61,12 +61,12 @@ func (e *MError) Error() string {
...
 
-func (e *MError) Unwrap() error {
-	return e
-}
-
 func (e *MError) Is(s error) bool {
-	return errors.Is(e.cerr, s)
+	is := errors.Is(e.cerr, s)
+	for _, e := range e.uerr {
+		is  = is || errors.Is(e, s)
```
djshow832, line 67: It can `break` if `is` is true.

xhebox: Done.

djshow832 (approved):

Issue #94, opened 2022-09-22
Dynamically update more configurations


djshow832:
Most of the configurations should be able to be updated dynamically and take effect immedicately.

PR #95, opened 2022-09-22
manager, config: watch and apply the latest log configuration online

djshow832:

Issue Number: ref #94

Problem Summary:
Log configurations need to take effect dynamically.

- Watch the update of the log configuration and apply them to the `LoggerManager`.
- Manage the `LoggerManager` by `Server` and remove the logs in `main()`.

TODO:
The `LoggerManager` relies on the `ConfigManager` to get the log path, but `ConfigManager` relies on `LoggerManager` to get the logger. Currently, `LoggerManager` logs to the path specified in the static config first and then switches to the path specified in the dynamic config. There are some ways to solve it:
- Forbid overwriting the log path
- Do not log in the `ConfigManager` before the logger is initialized
- There should be a global log path and a log path for each tenant. The global log path cannot be updated, but tenants can update theirs. Logging into different paths helps privilege control among tenants.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

```
$ echo '{"log-file": {"filename":"proxy.log"}}' | ./bin/tiproxyctl config log set
""
$ ./bin/tiproxyctl config log get
{"log-file":{"filename":"proxy.log"}}
$ cat proxy.log
[2022/09/23 10:27:05.069 +08:00] [INFO] [main.gin] [/api/admin/config/log] [status=200] [method=GET] [path=/api/admin/config/log] [query=] [ip=::1] [user-agent=Go-http-client/1.1] [latency=93.715µs]
```

Notable changes

- [x] Has weirctl change

### Release note

```release-note
None
```


pkg/manager/config/proxy.go
```diff
@@ -24,46 +24,154 @@ import (
...
-		}); err != nil {
+	for _, m := range e.metas {
+		if err := func(m imeta) error {
+			_, err := e.get(ctx, m.getPrefix(), "config")
+			if err != nil && errors.Is(err, ErrNoOrMultiResults) {
+				value, err := json.Marshal(m.getInitial(cfg))
+				if err != nil {
+					return err
+				}
+				if err = e.set(ctx, m.getPrefix(), "config", string(value)); err != nil {
```
xhebox, line 97: I want to change the prefix settings, it can be done in this PR, or another.

Previously, we do `/config/xxx/config`. I considered the suffix is for compatibility/reserving usages. But actually we could do `/config/v2/xxx` or `/v2/config/xxx` anyway. Let us just remove this useless suffix.

pkg/server/api/config.go
```diff
@@ -28,35 +28,45 @@ type configHttpHandler struct {
...
+	})
+	group.GET("/proxy", func(c *gin.Context) {
+		getConfig[config.ProxyServerOnline](h, c, mgrcfg.CfgServer)
+	})
+	group.PUT("/log", func(c *gin.Context) {
+		setConfig[config.LogOnline](h, c, mgrcfg.CfgLog)
+	})
+	group.GET("/log", func(c *gin.Context) {
+		getConfig[config.LogOnline](h, c, mgrcfg.CfgLog)
+	})
```
xhebox, line 71: I actually want something like `setConfig[config.ProxyServerOnline](h, c, mgrcfg) -> group.PUT(), gorup.GET()`.

But this LGTM, too. We can optimize it in another PR.

pkg/manager/config/proxy.go
```diff
@@ -24,46 +24,154 @@ import (
...
+func (e *ConfigManager) initMetas() {
+	e.metas = map[CfgType]imeta{
+		CfgServer: newMeta(pathPrefixProxyServer, func(cfg *config.Config) config.ProxyServerOnline {
+			return cfg.Proxy.ProxyServerOnline
+		}),
+		CfgLog: newMeta(pathPrefixLog, func(cfg *config.Config) config.LogOnline {
+			return cfg.Log.LogOnline
+		}),
+	}
+}
```
xhebox, line 86: I appreciate the usage of generic, but this does not feel necessary to me. I mean the current implementation is not as simple as non-generics while it is not as expressive as my expectation. The core is  `map[cfgType]imeta`, which is not generic. It basically means we can archive the same effect with:

```
&imetaStruct{
  prefix: "",
  cfgGetter: func() {
  },
  unmarshaler: func() {
  },
}
```

That is much easier to grasp than generics.

Generics should have a far more expressive power. But I don't see that: we still need a lot of manual getters, setters.`GetLogConfigWatch` is not replaced by `GetConfigWatch[T]` as my expectation. Even `GetConfig[T]` still need a explicity `cfgType`.

xhebox (commented):
Rest LGTM

pkg/manager/config/proxy.go
djshow832: I considered using a non-generic struct to implement the `imeta`. However, there will be some problems:
- Each struct needs to implement a `unmarshal` function
- Each struct needs to define a channel `chan XXX`

The usage of generic types in this file is complex, but it avoids duplicated code for each configuration type, assuming that there are more and more configurations will be updated online.

xhebox: > The usage of generic types in this file is complex, but it reduces duplicated code for each configuration type, assuming that there are more and more configurations will be updated online.

That is not true. Why not just define `chan` in `ConfigManager`? I can basically archive the same effect without `imeta`:

```
struct ConfigManager {
  xxChan
  yyChan
}
func (e *ConfigManager) watchCfgProxy[T OnlineCfgTypes](ctx context.Context, cfgType string, initial *T) (<-chan *T, error)
func (e *ConfigManager) SetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string, *T) error)
func (e *ConfigManager) GetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string) (*T, error)
func (e *ConfigManager) GetXXChan
func (e *ConfigManager) GetYYChan
```

Considering you need three lines for each imeta definition, one line of `e.xxChan, err = e.watchCfgProxy[config.LogOnline](ctx, "log", cfg.Proxy.ProxyServerOnline)` clearly wins. There is no need to define `GetProxyConfig`, or `SetProxyConfig` too. It is used only once anyway.

Or even more aggressive:

```
struct ConfigManager {
  GetXXChan func() <-chan *XX
  ...
}
e.GetXXChan, err = e.watchCfgProxy[config.XX](ctx, "xx", cfg.XX)
```

pkg/manager/config/proxy.go
```diff
@@ -24,46 +24,154 @@ import (
...
+
 func (e *ConfigManager) watchCfgProxy(ctx context.Context, cfg *config.Config) error {
```
xhebox, line 88: We should rename `watchCfgProxy` to something else like `watchConfigField`, `watchConfigItem`, or `watchConfig` maybe.

pkg/manager/config/proxy.go
djshow832: Do methods ( not functions) support generic types?

pkg/manager/config/proxy.go
djshow832: Maybe I need to also change the HTTP API from `config/proxy` to `config/server`. Does anything (such as operator) rely on this API currently?

pkg/manager/config/proxy.go
djshow832: > > The usage of generic types in this file is complex, but it reduces duplicated code for each configuration type, assuming that there are more and more configurations will be updated online.
> 
> That is not true. Why not just define `chan` in `ConfigManager`? I can basically archive the same effect without `imeta`:
> 
> ```
> struct ConfigManager {
>   xxChan
>   yyChan
> }
> func (e *ConfigManager) watchCfgProxy[T OnlineCfgTypes](ctx context.Context, cfgType string, initial *T) (<-chan *T, error)
> func (e *ConfigManager) SetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string, *T) error)
> func (e *ConfigManager) GetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string) (*T, error)
> func (e *ConfigManager) GetXXChan
> func (e *ConfigManager) GetYYChan
> ```
> 
> Considering you need three lines for each imeta definition, one line of `e.xxChan, err = e.watchCfgProxy[config.LogOnline](ctx, "log", cfg.Proxy.ProxyServerOnline)` clearly wins. There is no need to define `GetProxyConfig`, or `SetProxyConfig` too. It is used only once anyway.
> 
> Or even more aggressive:
> 
> ```
> struct ConfigManager {
>   GetXXChan func() <-chan *XX
>   ...
> }
> func (e *ConfigManager) watchCfgProxy[T OnlineCfgTypes](ctx context.Context, cfgType string, initial *T) (<-chan *T, error)
> func (e *ConfigManager) SetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string, *T) error)
> func (e *ConfigManager) GetConfig[T OnlineCfgTypes](ctx context.Context, cfgType string) (*T, error)
> 
> e.GetXXChan, err = e.watchCfgProxy[config.XX](ctx, "xx", cfg.XX)
> ```

Go doesn't support type parameters on methods. It only supports type parameters on functions.

I've replaced `CfgType` to `reflect.Type` so that the interfaces don't need an extra `CfgType` parameter.

pkg/manager/config/proxy.go
xhebox: It does use hard coded `API` path. Because `/lib/cli` does not support passing SSL, I have not actuall use `GetRootCmd`. Maybe change API path later.

But I think API names are OK, it is called `proxyServer` in config or so. Reasonable to call it `/config/proxy`.

lib/config/proxy.go
```diff
@@ -62,12 +62,16 @@ type Advance struct {
...
-type Log struct {
+type LogOnline struct {
 	Level   string  `yaml:"level,omitempty" toml:"level,omitempty" json:"level,omitempty"`
-	Encoder string  `yaml:"encoder,omitempty" toml:"encoder,omitempty" json:"encoder,omitempty"`
 	LogFile LogFile `yaml:"log-file,omitempty" toml:"log-file,omitempty" json:"log-file,omitempty"`
 }
 
+type Log struct {
+	Encoder string `yaml:"encoder,omitempty" toml:"encoder,omitempty" json:"encoder,omitempty"`
+	LogOnline
```
xhebox, line 72: I think you need to fix `LogOnline` instead of `ProxyServerOnline`...

xhebox (approved):
Rest LGTM

Issue #14, opened 2022-07-26
Simplified packetIO


xhebox:
- [ ] Without the complex logic of old weir(recoding DB/username, handle authentication by itself), it is completely safe to remove most code of `authenticator.go`. Also, it is more resonable to parse mysql in the net package, rather than in authentication package.
- [ ] Proxy protocol will be easier to implement on a clean packetIO. The reason is that we may need to act as a server for `PROXYV2` from cliens and act as a client to send `PROXYV2` to servers. Sadly we can not just import `blacktear23/go-proxyprotocol` because it only do parsing as a server.
  - #17
  - #25
  - #45
  - #96 more methods to packetIO
  - #113
  - testing more details
  - TLS has internal buffer, we should not buffer twice

PR #96, opened 2022-09-22
*: support proxy-protocol

xhebox:

Issue Number: ref #14, proxy section

Problem Summary: As title, it is coded as minimal as possible. One thing, it is closed by default: because tidb can not automatically check if incoming connections are proxy or not. (start tidb using tiup with proxy-protocol will fail the health check). That said we can only enable it if the upstream server supports.

1. misc changes in `conf/namespace/example.yaml`, `pkg/manager/router/router.go`, `Makefile`, `pkg/proxy/proxy.go`
2. `handshakeFirstTime` is rewritten such that handshakes between client and proxy is finished first. Another difference is that the new logic is more strict compare to the old logic, e.g. accurate `SSLRequest` are sent instead of `handshakeResp`.
3. `pkg/proxy/backend/util.go` a salt generator
4. fix a length bug in `ToBytes()`, also avoid panic if callers passed different address family.
5. set deadline for `pkg/proxy/net/packetio_test.go`
6. remove `weir` in github template

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
```
1. uncomment proxy-protocol in ./conf/proxy.yaml

2. set config for tidb(tiup ... --db.config ... or ./tidb-server -config ...)
[proxy-protocol]
networks = "*"
timeout = 10

3.  connect without failure
```
- [ ] No code

Notable changes

- [x] Has configuration change
- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


pkg/proxy/backend/authenticator.go
```diff
@@ -24,15 +24,14 @@ import (
...
-const supportedServerCapabilities = mysql.ClientLongPassword | mysql.ClientFoundRows | mysql.ClientLongFlag |
-	mysql.ClientConnectWithDB | mysql.ClientNoSchema | mysql.ClientODBC | mysql.ClientLocalFiles | mysql.ClientIgnoreSpace |
-	mysql.ClientProtocol41 | mysql.ClientInteractive | mysql.ClientSSL | mysql.ClientIgnoreSigpipe |
-	mysql.ClientTransactions | mysql.ClientReserved | mysql.ClientSecureConnection | mysql.ClientMultiStatements |
-	mysql.ClientMultiResults | mysql.ClientPluginAuth | mysql.ClientConnectAtts | mysql.ClientPluginAuthLenencClientData |
-	mysql.ClientDeprecateEOF
+const supportedServerCapabilities = pnet.ClientLongPassword | pnet.ClientFoundRows | pnet.ClientLongFlag |
+	pnet.ClientConnectWithDB | pnet.ClientLocalFiles | pnet.ClientProtocol41 | pnet.ClientInteractive | pnet.ClientSSL |
+	pnet.ClientTransactions | pnet.ClientSecureConnection | pnet.ClientMultiStatements |
+	pnet.ClientMultiResults | pnet.ClientPluginAuth | pnet.ClientConnectAttrs | pnet.ClientDeprecateEOF
```
xhebox, line 34: The old `supportedServerCapabilities` is a fullset for testing. The new capabilities are aligned to the current version of tidb.

djshow832: This is a set of supported capabilities, not default capabilities. If the client contains other capabilities, the server will report an error.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,116 @@ import (
...
+			return errors.WithStack(err)
+		}
+		if isSSL {
+			return errors.WithStack(errors.New("expect handshake resp"))
+		}
+	} else {
+		binary.LittleEndian.PutUint32(pkt, (clientCapability | pnet.ClientSSL).Uint32())
+	}
+	if cc := clientCapability & serverCapability; clientCapability^cc != 0 {
+		logger.Info("unsupported capability from the frontend", zap.Stringer("common", cc), zap.Stringer("frontend", clientCapability^cc), zap.Stringer("proxy", serverCapability^cc))
```
djshow832, line 91: This is not right. Some capabilities are not really capabilities, they are just client options.
E.g. the default server capability doesn't contain `ClientNoSchema` but the client may send it to specify that the handshake response doesn't contain a schema. The server still supports it.
That's why I added a `supportedServerCapabilities` to specify what the proxy supports.

xhebox: Then it is conflicted with the official mysql document:

> The client should only announce the capabilities in the [Protocol::HandshakeResponse](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake_response.html) that it has in common with the server.

Server should at least warn about uncommon flags. Capability negotiation works by assuming common capabilities.

TBH, I don't think it is good idea to be slient about those unsupported capabilities. You can see that TiDB does not support `ClientNoSchema`, too. But TiDB will be slient about it, which means it only works if clients respect the mysql protocol strictly.

```
[xhe@M1PRO a]$ rg ClientNoSchema ../pingcap/tidb
../pingcap/tidb/parser/mysql/const.go
140:	ClientNoSchema

No other occurrences
```

djshow832: The MySQL document is not always right. See  https://github.com/pingcap/tidb/blob/master/server/conn.go#L500

djshow832: Besides, the proxy can support `CLIENT_ODBC` because it has no special behaviors.
 https://dev.mysql.com/doc/internals/en/capability-flags.html#flag-CLIENT_ODBC

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,116 @@ import (
...
-	serverPkt, serverCapability, err := auth.readInitialHandshake(backendIO)
-	if serverPkt != nil {
-		if writeErr := clientIO.WritePacket(serverPkt, true); writeErr != nil {
-			return writeErr
+
+	serverCapability := auth.supportedServerCapabilities
+	if frontendTLSConfig == nil {
+		serverCapability ^= pnet.ClientSSL
+	}
+	if err := clientIO.WriteInitialHandshake(serverCapability.Uint32(), salt, mysql.AuthCachingSha2Password); err != nil {
```
djshow832, line 65: Why does the proxy send a handshake packet first? I read the proxy protocol and I think the proxy can just forward the handshake from the server to the client.

pkg/proxy/net/proxy.go
```diff
@@ -81,69 +85,72 @@ type Proxy struct {
...
 }
 
-func (p *Proxy) ToBytes() []byte {
-	buf := make([]byte, len(proxyV2Magic)+4)
-	ptr := copy(buf, proxyV2Magic)
+func (p *Proxy) ToBytes() ([]byte, error) {
+	magicLen := len(proxyV2Magic)
+	buf := make([]byte, magicLen+4)
+	_ = copy(buf, proxyV2Magic)
+	buf[magicLen] = byte(p.Version<<4) | byte(p.Command&0xF)
```
djshow832, line 92: The command part should always be `0x01`, otherwise, the server will read the address of the proxy.
See https://github.com/blacktear23/go-proxyprotocol/blob/master/proxy_protocol.go#L311-L313

xhebox: That is controlled by the external part. This is just a correct library. And in the current implementation, it is always `PROXY`. proxy sould also handle `LOCAL`, but that will not be done in this PR.

pkg/proxy/backend/authenticator.go
xhebox: > Besides, the proxy can support CLIENT_ODBC because it has no special behaviors.

Then we should add it to the supported set.

> The MySQL document is not always right. 

It is that mysql-client is wrong. We should respect the spec as much as possible.

>  You cannot test all clients.

If you feel it is too strict to forbid all cases. We can just report the difference without really disconnect it.

pkg/proxy/backend/authenticator.go
djshow832: > some clients does not really respect the standard

Does it include AWS NLB?

> But some servers won't read the header lazily.

We only need to support TiDB. If TiDB supports reading lazily, then the proxy can connect to TiDB first and read the initial handshake.
TiProxy will never support other servers, because it depends on PD, `SHOW SESSION_STATES`, token based auth, etc.

> Of course, another reason is that per-namespace and other refactors relies on such behaviors.

We should respect how gateway routes connections. I was told it's an L4 proxy, so I don't think it generates an initial handshake by itself.

The reason why the proxy should not write the initial handshake packet by itself is that the handshake packet sent by the server may change, it includes salt, capability, and default auth plugin.

pkg/proxy/net/proxy.go
djshow832: > If tiproxy forward LOCAL to tidb, it is also correct. That means another proxy wants to do health checking for TiDB through tiproxy.

Is it the only case when another proxy sends `LOCAL`? How did you get this information?

> Even if you rewrite the LOCAL to PROXY, you probably will get the address of the uppper proxy instead of a mysql client.

Why? Both TiProxy and TiDB will get the address of the upstream sender.

xhebox: > Is it the only case when another proxy sends LOCAL? How did you get this information?

The standard defines as so. If we want to enforce such behaviors, we could reject handshake for `LOCAL` command on tiproxy or tidb but still being responsive.

> Why? Both TiProxy and TiDB will get the address of the upstream sender.

Because `LOCAL` means the upstream sender is a proxy, instead of a client.

pkg/proxy/backend/authenticator.go
xhebox: > Does it include AWS NLB?

Yes.

> TiProxy will never support other servers, because it depends on PD, SHOW SESSION_STATES, token based auth, etc.

It could remove dependencies on PD, if dev tier GW and tiproxy merges as one. BTW, tiproxy could be used as a proxy in multi layer proxies. There could be lots of cases where tiproxy does not connect tidb directly.

> We should respect how gateway routes connections. I was told it's an L4 proxy.

Parsing and intercepting mysql proxy means it is already an L7 proxy. Can we remove `packetIO` from this package? Or can we do not migrate session by sending SQL that is not from clients? It is not even a L4 proxy in the respect of mysql protocol.

Also, handshaking clients first are the only way to do per-namespace routing, unless we route connections by public IP only.

> The reason why the proxy should not write the initial handshake packet by itself is that the handshake packet sent by the server may change, it includes salt, capability, and default auth plugin.

1. `capability` does matter, but not much. Most packets are intercepted by tiproxy, which means we could handle the case of different capabilities. Also, since we support same capabilities as tidb, there is no problem for now.
2. `auth` will be handled by `switch auth request`
3. `salt` or `plugin-auth-data` could be improved by caching. The next time handshake from the same user would know which backend to connect. In this case, we can forward initial handshake from backends directly. This has beyond the scope of this PR.

If you feel handshaking first is too dangerous, I could revert it in this PR: forward server `initialHandshake` in this PR. And do `handshake clients first + client/backend caching` in another PR.

pkg/proxy/net/proxy.go
djshow832:
```
LOCAL : the connection was established on purpose by the proxy
    without being relayed. 
PROXY : the connection was established on behalf of another node.
```

If a proxy sends `LOCAL` and TiProxy relays the message, then TiProxy should send `PROXY`, because it was established on behalf of another node.

pkg/proxy/backend/authenticator.go
djshow832: > salt or plugin-auth-data could be improved by caching.

I didn't mean the performance. The salt generated from the TiDB each time is different. TiProxy cannot generate a random salt if it doesn't authenticate the client by itself.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,123 @@ import (
...
+	_, backendCapabilityU, err := auth.readInitialHandshake(backendIO)
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
-	if serverCapability&mysql.ClientSSL == 0 {
+	backendCapability := pnet.Capability(backendCapabilityU)
+	if common := commonCaps & backendCapability; (commonCaps^common)&^pnet.ClientSSL != 0 {
+		logger.Info("backend does not support capabilities from client&proxy", zap.Stringer("common", common), zap.Stringer("server", commonCaps^common), zap.Stringer("backend", backendCapability^common))
+	}
```
djshow832, line 133: This is not right, considering `ClientPluginAuthLenencClientData` and other capabilities that are in default backend capabilities.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,123 @@ import (
...
-		// Rewrite the packet with ClientSSL enabled because we always connect to TiDB with TLS.
-		pktWithSSL := make([]byte, len(clientPkt))
-		pnet.DumpUint16(pktWithSSL[:0], clientCapability|uint16(mysql.ClientSSL))
-		copy(pktWithSSL[2:], clientPkt[2:])
-		clientPkt = pktWithSSL
-	}
-	if err = backendIO.WritePacket(clientPkt, true); err != nil {
-		return err
+	// write SSL Packet
+	if err := backendIO.WritePacket(pkt[:32], true); err != nil {
```
djshow832, line 141: The backend must report an error because the salts don't match.

xhebox: This is a subproblem of handshaking.

pkg/proxy/backend/authenticator.go
xhebox: I don't get it. If `ClientPluginAuthLenencClientData` is supported, it should be sent by the backend.

pkg/proxy/net/proxy.go
xhebox: Makes sense, I could do the change. I think the core point is that `LOCAL` does not care about the original IP, but anyway.

pkg/proxy/backend/authenticator.go
xhebox: > I didn't mean the performance. The salt generated from the TiDB each time is different. TiProxy cannot generate a random salt if it doesn't authenticate the client by itself.

I didn't mean performance too. Caching will enable us to choose the correct namespace for routing by pairing IP and other information. In other words, it gives a chance to just forward server initial handshake.

xhebox: Done, forwarding initial handshake first.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -87,9 +88,10 @@ type BackendConnManager struct {
 func NewBackendConnManager(logger *zap.Logger, connectionID uint64) *BackendConnManager {
 	return &BackendConnManager{
 		logger:         logger,
+		salt:           GenerateSalt(20),
```
djshow832, line 91: We can remove this salt part currently.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,134 @@ import (
...
+	if frontendCapability^commonCaps != 0 {
+		logger.Error("frontend send capabilities unsupported by proxy", zap.Stringer("common", commonCaps), zap.Stringer("frontend", frontendCapability^commonCaps), zap.Stringer("proxy", proxyCapability^commonCaps))
+		return errors.Wrapf(ErrCapabilityNegotiation, "%s", frontendCapability^commonCaps)
+	}
+	resp := pnet.ParseHandshakeResponse(pkt)
+	auth.capability = commonCaps.Uint32()
+	auth.user = resp.User
+	auth.dbname = resp.DB
+	auth.collation = resp.Collation
+	auth.attrs = resp.Attrs
```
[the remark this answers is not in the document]
xhebox: It will be even longer later. I can extract something out, but I think it is not time.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,134 @@ import (
...
-		clientPkt = pktWithSSL
+		binary.LittleEndian.PutUint32(pkt, (frontendCapability | pnet.ClientSSL).Uint32())
 	}
-	if err = backendIO.WritePacket(clientPkt, true); err != nil {
-		return err
+	if commonCaps := frontendCapability & requiredCapabilities; commonCaps != requiredCapabilities {
+		logger.Error("require frontend capabilities", zap.Stringer("common", commonCaps), zap.Stringer("required", requiredCapabilities))
+		return errors.Wrapf(ErrCapabilityNegotiation, "require %s", requiredCapabilities&^commonCaps)
+	}
+	commonCaps := frontendCapability & proxyCapability
```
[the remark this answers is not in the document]
xhebox: This should be investigated, I rememered mysql-server sometimes will only use the first capabilities.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,134 @@ import (
...
-	mysql.ClientTransactions | mysql.ClientReserved | mysql.ClientSecureConnection | mysql.ClientMultiStatements |
-	mysql.ClientMultiResults | mysql.ClientPluginAuth | mysql.ClientConnectAtts | mysql.ClientPluginAuthLenencClientData |
-	mysql.ClientDeprecateEOF
+const requiredCapabilities = pnet.ClientProtocol41
+const supportedServerCapabilities = pnet.ClientLongPassword | pnet.ClientFoundRows | pnet.ClientLongFlag |
+	pnet.ClientConnectWithDB | pnet.ClientNoSchema | pnet.ClientODBC | pnet.ClientLocalFiles | pnet.ClientIgnoreSpace |
+	pnet.ClientInteractive | pnet.ClientSSL | pnet.ClientIgnoreSigpipe |
+	pnet.ClientTransactions | pnet.ClientReserved | pnet.ClientSecureConnection | pnet.ClientMultiStatements |
+	pnet.ClientMultiResults | pnet.ClientPluginAuth | pnet.ClientConnectAttrs | pnet.ClientPluginAuthLenencClientData |
+	pnet.ClientDeprecateEOF | requiredCapabilities
```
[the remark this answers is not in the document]
xhebox: Then I can investigate and see if TiDB really supports them.

pkg/proxy/backend/backend_conn_mgr.go
xhebox: OK, I can revert this for now.

pkg/proxy/backend/authenticator.go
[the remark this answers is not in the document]
djshow832: > it is not time

Do you mean next time?

pkg/proxy/backend/authenticator.go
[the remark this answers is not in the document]
xhebox: The answer is `NO`. TiDB does not support these capbilities. So we should not include them here. If clients assume that TiDB is actually peform as if they are supported, something may go wrong.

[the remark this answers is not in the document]
xhebox: No, neither TiDB nor tiproxy support these caps. It is wrong to send them to clients. We should not include them in the supported set.

pkg/proxy/backend/authenticator.go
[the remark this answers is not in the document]
xhebox: Confirmed, It is same in my several tests. The behavior should be determined by `libmysql` linked to sysbench. Sysbench it self does not control this.

pkg/proxy/backend/authenticator.go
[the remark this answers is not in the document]
xhebox: Yes.

pkg/proxy/backend/authenticator.go
```diff
@@ -24,74 +24,134 @@ import (
...
+	proxyCapability := auth.supportedServerCapabilities
+	if frontendTLSConfig == nil {
+		proxyCapability ^= pnet.ClientSSL
 	}
+
+	// read backend initial handshake
+	backendHandshake, backendCapabilityU, err := auth.readInitialHandshake(backendIO)
 	if err != nil {
-		return err
+		return errors.WithStack(err)
```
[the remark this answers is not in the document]
xhebox: Done for this function.

pkg/proxy/net/packetio_mysql.go
```diff
@@ -75,22 +86,28 @@ func (p *PacketIO) WriteShaCommand() error {
...
-		return nil, errors.WithStack(errors.Errorf("%w: but got less than 2 bytes", ErrExpectSSLRequest))
+	if len(pkt) < 32 {
+		err = errors.WithStack(errors.Errorf("%w: but got less than 32 bytes", ErrExpectSSLRequest))
+		return
 	}
 
-	capability := uint32(binary.LittleEndian.Uint16(pkt[:2]))
-	if capability&mysql.ClientSSL == 0 {
-		return nil, errors.WithStack(errors.Errorf("%w: but capability flags has no SSL", ErrExpectSSLRequest))
+	isSSL = len(pkt) <= 32
```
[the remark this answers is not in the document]
xhebox: Done.

pkg/proxy/backend/util.go
```diff
@@ -0,0 +1,33 @@
...
+package backend
+
+import _ "unsafe"
+
+//go:linkname Uint32N runtime.fastrandn
+func Uint32N(a uint64) uint64
+
+// Buf generates a random string using ASCII characters but avoid separator character.
+// Ref https://github.com/mysql/mysql-server/blob/5.7/mysys_ssl/crypt_genhash_impl.cc#L435.
+func GenerateSalt(size int) []byte {
```
djshow832, line 24: We don't need to generate real salt. We can send any salts and then switch auth-plugin.

pkg/proxy/client/client_conn.go
```diff
@@ -40,24 +40,25 @@ type ClientConnection struct {
 	nsmgr             *namespace.NamespaceManager
 	ns                *namespace.Namespace
 	connMgr           *backend.BackendConnManager
+	proxyProtocol     bool
 }
-
-func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager) *ClientConnection {
-	pkt := pnet.NewPacketIOWrapErr(conn, ErrClientConn)
+func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager, proxyProtocol bool) *ClientConnection {
```
djshow832, line 45: Leave a blank line before the function.

pkg/proxy/client/client_conn.go
```diff
@@ -69,22 +70,22 @@ func (cc *ClientConnection) connectBackend(ctx context.Context) error {
...
-		cc.logger.Info("new connection fails", zap.String("remoteAddr", cc.Addr()), zap.Error(err))
+		cc.logger.Info("new connection fails", zap.Error(err))
 		return
 	}
 
 	if err := cc.processMsg(ctx); err != nil {
 		clientErr := errors.Is(err, ErrClientConn)
 		if !(clientErr && errors.Is(err, io.EOF)) {
-			cc.logger.Info("process message fails", zap.String("remoteAddr", cc.Addr()), zap.Error(err), zap.Bool("clientErr", clientErr), zap.Bool("serverErr", !clientErr))
+			cc.logger.Info("process message fails", zap.Error(err), zap.Bool("clientErr", clientErr), zap.Bool("serverErr", !clientErr))
```
djshow832, line 88: You can assign the address in the proxy protocol to the ClientConnection so that it can log client addresses.
Or you can use `logger.With` to contain the real client address.

pkg/proxy/client/client_conn.go
```diff
@@ -40,24 +40,25 @@ type ClientConnection struct {
...
 	connMgr           *backend.BackendConnManager
+	proxyProtocol     bool
 }
-
-func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager) *ClientConnection {
-	pkt := pnet.NewPacketIOWrapErr(conn, ErrClientConn)
+func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager, proxyProtocol bool) *ClientConnection {
+	opts := make([]pnet.PacketIOption, 0, 1)
+	if proxyProtocol {
+		opts = append(opts, pnet.WithProxy, pnet.WithClient)
```
djshow832, line 48: Never used?

pkg/proxy/client/client_conn.go
xhebox: I have done that, check `pkg/proxy/proxy.go`.

djshow832:
HandshakeSecondTime also needs proxy protocol. This requires the authenticator to keep the real client address. You can do it in another PR.

pkg/proxy/backend/util.go
xhebox: It is already modified server initial handshake forwarding. A complete solution will come in the next PR. But anyway, I am going to delete this.

pkg/proxy/client/client_conn.go
xhebox: I guess something is wrong when merging master...

xhebox:
> HandshakeSecondTime also needs proxy protocol. This requires the authenticator to keep the real client address. You can do it in another PR.

I agree. We have discussed too much on this PR.

pkg/proxy/client/client_conn.go
```diff
@@ -40,24 +40,26 @@ type ClientConnection struct {
 	nsmgr             *namespace.NamespaceManager
 	ns                *namespace.Namespace
 	connMgr           *backend.BackendConnManager
+	proxyProtocol     bool
 }
 
-func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager) *ClientConnection {
-	pkt := pnet.NewPacketIOWrapErr(conn, ErrClientConn)
+func NewClientConnection(logger *zap.Logger, conn net.Conn, frontendTLSConfig *tls.Config, backendTLSConfig *tls.Config, nsmgr *namespace.NamespaceManager, bemgr *backend.BackendConnManager, proxyProtocol bool) *ClientConnection {
+	opts := make([]pnet.PacketIOption, 0, 1)
```
djshow832, line 47: Maybe make the cap as 2?

pkg/proxy/proxy.go
```diff
@@ -61,6 +62,7 @@ func NewSQLServer(logger *zap.Logger, cfg config.ProxyServer, scfg config.Securi
 		mu: serverState{
 			tcpKeepAlive:   cfg.TCPKeepAlive,
 			maxConnections: cfg.MaxConnections,
+			proxyProtocol:  cfg.ProxyProtocol != "",
```
djshow832, line 65: It's not updated in `SQLServer.Run`.

xhebox: Done.

djshow832 (approved):

xhebox:
`PSMultiresult` tests are removed. Since proxy can not handle this and won't send this, it does not matter backend/frontend has this capability flag or not.

PR #98, opened 2022-09-26
*: ctl TLS refine

xhebox:

Issue Number: ref #69

Problem Summary: The main purpose is to enable passing TLSConfig programatically, which makes it embeddable in other applications like operator.

1. `GetRootCmd` now accepts `tls.Config`, the main purpose
2. Silent usage on errors. Only manually print usages on argument mismatch or so.
3. Support passing tls set from cli, and automatically request with or without SSL
4. Fix handling of `res.StatusCode != http.StatusOK`
5. Handle cases where server with TLS but clients request plain http and vice versa

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
Just enable server with or with TLS and test

One special case is to test !200 cases. Start tiproxy without PD, and you will hit `service not ready` in the first 5 seconds or so 100%.
```
- [ ] No code

Notable changes

- [x] Has weirctl change

### Release note

```release-note
None
```

### DIFF
--- a/lib/cli/main.go
+++ b/lib/cli/main.go
@@ -15,16 +15,20 @@
 package cli
 
 import (
+	"crypto/tls"
 	"net/http"
 
+	"github.com/pingcap/TiProxy/lib/config"
+	"github.com/pingcap/TiProxy/lib/util/security"
 	"github.com/spf13/cobra"
 	"go.uber.org/zap"
 )
 
-func GetRootCmd() *cobra.Command {
+func GetRootCmd(tlsConfig *tls.Config) *cobra.Command {
 	rootCmd := &cobra.Command{
-		Use:   "tiproxyctl",
-		Short: "cli",
+		Use:          "tiproxyctl",
+		Short:        "cli",
+		SilenceUsage: true,
 	}
 
 	ctx := &Context{}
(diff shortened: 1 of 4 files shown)


lib/cli/main.go
```diff
@@ -15,23 +15,30 @@
...
+		Short:        "cli",
+		SilenceUsage: true,
 	}
 
 	ctx := &Context{}
 
 	curls := rootCmd.PersistentFlags().StringArray("curls", []string{"localhost:3080"}, "API gateway addresses")
 	logEncoder := rootCmd.PersistentFlags().String("log_encoder", "tidb", "log in format of tidb, console, or json")
 	logLevel := rootCmd.PersistentFlags().String("log_level", "info", "log level")
+	caPath := rootCmd.PersistentFlags().String("ca", "", "CA to verify server certificates, set to 'skip' if want to enable SSL without verification")
```
djshow832, line 39: Why need to set `skip`? Why not set it to an empty string or skip setting it?

xhebox: 1. `""` means no TLS, retry will use the system CA
2. `skip` means TLS without CA, fallback to HTTP if server may be HTTP
3. otherwise, it is normal TLS with specified CA, fallback to HTTP if server may be HTTP

So, in short, `skip` is like `curl -k`, give a chance to use TLS insecurely(for example testing).

djshow832: You have `certPath` and `keyPath`. If `certPath` and `keyPath` are specified, then TLS is enabled without verification. This is what TiDB does in the SQL port.

xhebox: That is not how tiproxy works. I invoked `BuildClientTLSConfig`, only CA or skip-ca will enable tls. Certs are ignored without ca.

PR #104, opened 2022-09-30
*: improve docker scripts

xhebox:

Issue Number: close None

Problem Summary: as title.

1. shrink docker image size
2. add `make docker-release` using `docker buildx` to push multiple-platform iamges
3. continue to improve CI hit things: always fully update cache after merging, only download cache for PRs. It is actually harder than expected to archive the effect.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


.dockerignore
```diff
@@ -1 +0,0 @@
-.gitignore
```
djshow832, line 1: What about this?
```
!**/*.go
!**/*.mod
!**/*.sum
```

xhebox: It means source files will not be included. But the docker script will actually build binary in the docker, so source files are needed.

xhebox: I think we need `Makefile`, too. We need to keep adding new source file types into the list. It makes no difference with adding new ignoring file types.

djshow832 (approved):

PR #112, opened 2022-10-11
*: support clustering

xhebox:

Issue Number: ref #69

Problem Summary: Support a cluster of tiproxy nodes.

What is changed and how it works: Add five new cmd flags.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
```
n1: 3080,3081,6000
n2: 4080,4081,7000

 rm -rf work && ./bin/tiproxy --node_name=n1 --bootstrap_clusters "n2=127.0.0.1:4081" --config conf/proxy.yaml
rm -rf work2 && ./bin/tiproxy --node_name=n2 --bootstrap_clusters "n1=127.0.0.1:3081" --config conf2/proxy.yaml

After bootstraping

stop one node and restart it by ./bin/tiproxy  --config conf/proxy.yaml
```
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -19,10 +19,10 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
 
 	ginzap "github.com/gin-contrib/zap"
 	"github.com/gin-gonic/gin"
@@ -257,7 +257,11 @@ func buildEtcd(cfg *config.Config, logger *zap.Logger, engine *gin.Engine) (srv 
 		return nil, errors.New("cluster_name can not be empty")
 	}
 	if cfg.Cluster.NodeName == "" {
-		cfg.Cluster.NodeName = fmt.Sprintf("%s-%d", cfg.Cluster.ClusterName, time.Now().UnixMicro())
+		hname, err := os.Hostname()
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		cfg.Cluster.NodeName = fmt.Sprintf("%s-%s", cfg.Cluster.ClusterName, hname)
 	}
 
 	cnt := 0
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	ginzap "github.com/gin-contrib/zap"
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -251,7 +252,28 @@ func (s *Server) Close() error {
...
+	}
+	if cfg.Cluster.BootstrapDurl != "" {
+		cnt++
+	}
+	if cfg.Cluster.BootstrapDdns != "" {
+		cnt++
+	}
+	if cnt > 1 {
+		return nil, errors.New("you can only pass one 'bootstrap_xxx' to bootstrap the node, or leave them empty to start a single-node cluster")
+	}
```
djshow832, line 275: Why make it so complicated, are other components also like this?

xhebox: This is common for etcd and other cloud native applications. Other pingcap components relies on the central PD(they only need PD).

PD only provide `bootstrapClusters` options, and a sugar option of `bootstrapCluster` by `-join xxx:2380` if you don't want to write all clusters by hands. I have no plan to include that sugar option in this PR.

Honestly, discovery or dns is more convenient.

To use `--initial-clusters`, you need to specify all nodes. To use `-join`, the cluster must be bootstraped already. Whatever you need to pass different options fot different PD and probably follow some orders.

Operator has implemented some very complex code.

djshow832: In my option, there's even no need to merge this PR, right? Proxies don't need to know each other, except for updating configurations. But updating configurations can be achieved by configuring the etcd address.

pkg/server/server.go
```diff
@@ -251,7 +252,28 @@ func (s *Server) Close() error {
 	return errors.Collect(ErrCloseServer, errs...)
 }
 
-func buildEtcd(cfg *config.Config, logger *zap.Logger, pubAddr string, engine *gin.Engine) (srv *embed.Etcd, err error) {
+func buildEtcd(cfg *config.Config, logger *zap.Logger, engine *gin.Engine) (srv *embed.Etcd, err error) {
+	if cfg.Cluster.ClusterName == "" {
+		return nil, errors.New("cluster_name can not be empty")
+	}
+	if cfg.Cluster.NodeName == "" {
+		cfg.Cluster.NodeName = fmt.Sprintf("%s-%d", cfg.Cluster.ClusterName, time.Now().UnixMicro())
```
djshow832, line 260: Does PD also generate it like this?

xhebox: No, but in a smilar fashion. It needs to be unique in the cluster.

pkg/server/server.go
xhebox: Only if you want to remove etcd from tiproxy, and relies on an external ETCD. I think it makes things complicated. If you have PD, you don't want another etcd. If you use tiproxy with static servers, you need another etcd.

pkg/server/server.go
djshow832: Can we also generate it like that? A hostname is better than a random timestamp.

djshow832 (approved):

Issue #15, opened 2022-07-26
SSL SNI routing


xhebox:
- [ ] Without storing username and password, proxy can not complete the authentication without real backend servers. Proxy needs to fake handshake and response for both clients and servers. But if we route by SNI, we only need to fake `inital handshake` for clients and parse `SSL request` from clients. It greatly simplify the parsing of mysql protocols, leaving clients to real servers. 
- [ ] This can remove literally lots of code of the complex frontend/backend logic.
  - #16 preparation, merged into #18
  - #18 preparation step two
  - #27

PR #119, opened 2022-10-20
*: fix static backends & support route by user

xhebox:

Issue Number: ref #15

Problem Summary: Minimal effort to support multi-namespace routing by user. Also fixed broken static routing. It is initially coded for hackathon, but mergeable anyway.

It also reveals  some problems:

1. The abstraction `client -> backend -> auth` is a little bit redundant, or too ideal: we need to pass options back and forth. In fact, we only need one class to handle connection with seperated files for readability.
2. It can not be well tested with the current `mockProxy`: `mp` will construct a fixed backend with `connectionID = 0, proxy = false, require-backend-tls = false` and also `namespaceManager = nil` now.
We probably will add more and more options. Yet `clientConn` and `proxyServer` is not tested. Need something more flexible than `mp/mb/mc`: a better API easier for testing instead of more helpers to target different cases.

The next step would be PD and TLS per namespace. Then we can possibly combine client/backend to enable more unit testing on these codes.

1. Add `GetNamespaceByUser` and `getBackendIO` hook.
2. Handle EOF in both handshake and packet relay in `clientConn`.
3. Do not fail it no instances are specified for the static backend. Instead, report errors on connection.
4. Check health first for static backends, then sleep.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
Comment PD addrs to enable static backends first.
```
tiproxyctl namespace import conf/namespace
tiproxyctl namespace commit default ns2

mycli -u xhe -P 6000
mycli -u root -P 6000
```
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/manager/namespace/manager.go
+++ b/pkg/manager/namespace/manager.go
@@ -48,6 +48,7 @@ func (mgr *NamespaceManager) buildNamespace(cfg *config.Namespace) (*Namespace, 
 	}
 	return &Namespace{
 		name:   cfg.Namespace,
+		user:   cfg.Frontend.User,
 		router: rt,
 	}, nil
 }
@@ -95,6 +96,18 @@ func (n *NamespaceManager) GetNamespace(nm string) (*Namespace, bool) {
 
 	ns, ok := n.nsm[nm]
 	return ns, ok
+}
+
+func (n *NamespaceManager) GetNamespaceByUser(user string) (*Namespace, bool) {
+	n.RLock()
+	defer n.RUnlock()
+
+	for _, ns := range n.nsm {
+		if ns.User() == user {
+			return ns, true
+		}
+	}
+	return nil, false
 }
 
 func (n *NamespaceManager) RedirectConnections() []error {
--- a/conf/namespace/test.yaml
+++ b/conf/namespace/test.yaml
@@ -1,0 +1,9 @@
+namespace: "ns2"
+frontend:
+  user: "xhe"
+  security:
+backend:
+  instances:
+    - "127.0.0.1:7000"
+  selector-type: "random"
+  security:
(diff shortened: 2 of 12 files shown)


pkg/manager/namespace/manager.go
```diff
@@ -97,6 +98,18 @@ func (n *NamespaceManager) GetNamespace(nm string) (*Namespace, bool) {
...
+func (n *NamespaceManager) GetNamespaceByUser(user string) (*Namespace, bool) {
+	n.RLock()
+	defer n.RUnlock()
+
+	for _, ns := range n.nsm {
+		if ns.User() == user {
+			return ns, true
+		}
+	}
+	return nil, false
```
djshow832, line 110: I'm afraid distinguishing namespaces by usernames may never be applied. If you have listened to the share from the PM, creating different usernames may still be trouble for users.
For the dev tier, they are getting the tenant information from the user names. It's different with this way.

djshow832: But I think it's fine to support it now. We may remove it in the future if it's unncessary.

xhebox: We can go on development on routing while not apply any FD for official releases. But it is natural to develop: this PR merely has 100 LOC. I am still willing to do more refactors to make the code hackable and easier, both for dev and testing.

As for the dev tier, it is same, they are getting tenant information from user names, so does me. It is more of a problem of how to integrate GW and tiproxy.

xhebox: We can remove routing by user anytime and we can add it back anytime. I want such a framework instead of a fixed framework for "the current specialized product". So i added `getBackendIO` to give more flexibility. This it the first motivation.

Also PMs clearly did not have enough energy to cover all aspects of products. A flexible and high-performance GW is absolutely a possible HTAP solution. According to Liu Tang, the reason they rejected "one endpoint solution" is that it is hard to implement and unknown what to do. But from this PR, it may not be that hard and it can be very clear.

djshow832 (approved):

PR #128, opened 2022-11-02
*: fix TLS cases

xhebox:

Issue Number: close None

Problem Summary: If server enables TLS, but we don't have certs to complete TLS handshake, there is a panic.

1. check `backendTLSConfig`
2. check `GetClientCertificate`

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/manager/cert/cert_info.go
+++ b/pkg/manager/cert/cert_info.go
@@ -26,6 +26,8 @@ import (
 	"go.uber.org/zap"
 )
 
+var emptyCert = new(tls.Certificate)
+
 // Security configurations don't support dynamically updating now.
 type certInfo struct {
 	cfg         config.TLSConfig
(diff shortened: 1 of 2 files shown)


djshow832 (approved):

pkg/manager/cert/cert_info.go
```diff
@@ -124,7 +126,12 @@ func (ci *certInfo) customizeTLSConfig(tlsConfig *tls.Config) *tls.Config {
 		}
 	} else {
 		tlsConfig.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
-			return ci.getCertificate(), nil
+			cert := ci.getCertificate()
```
djshow832, line 129: Why not return the `emptyCert` in the `getCertificate`? It will also work when `ci.isServer==true`.

pkg/manager/cert/cert_info.go
```diff
@@ -124,7 +126,12 @@ func (ci *certInfo) customizeTLSConfig(tlsConfig *tls.Config) *tls.Config {
...
+			if cert == nil {
```
djshow832, line 130: When is `cert == nil`? If the tlsConfig is nil, it won't come here. If the tlsConfig is not nil, can `cert` be nil?

xhebox: A tls config with only CA, or even without CA `insecureSkipVerify`.

pkg/manager/cert/cert_info.go
xhebox: `GetCertificate` and `GetClientCertificate` has different requirements. I mean `GetCertificate` should return nil.

PR #160, opened 2022-12-26
backend: remove getBackendIO

xhebox:

Issue Number: close None

Problem Summary: Another refactor to clean the code. `getBackendIO` is added to do namespace routing. Now it is done in `HandshakeHandler` by returning the correct router. So it is completely fine to remove `getBackendIO` now.

1. Split `ScoreBasedRouter` into a separate file.
2. Add `StaticRouter` for tests.
3. Add `ConnEventReceiver` interface as parts of `Router`
4. `getBackendIO` is moved into `handshakeFirstTime`
5. `handshakeFirstTime` is now a method of `BackendConnMgr` due to dependency problems.(handler is stored in the manager, not the authenticator)
6. Replace `github.com/pingcap/tidb/util.WaitGroupWrapper` with `github.com/pingcap/TiProxy/lib/util/waitgroup.WaitGroup`
7. `CustomHandshakeHandler` moved into `handshake_handler.go`. Minor refactored.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/proxy/backend/authenticator.go
+++ b/pkg/proxy/backend/authenticator.go
@@ -106,7 +106,6 @@ type backendIOGetter func(ctx ConnContext, auth *Authenticator, resp *pnet.Hands
 func (auth *Authenticator) handshakeFirstTime(logger *zap.Logger, clientIO, backendIO *pnet.PacketIO, handshakeHandler HandshakeHandler,
 	getBackendIO backendIOGetter, frontendTLSConfig, backendTLSConfig *tls.Config) error {
 	clientIO.ResetSequence()
-	auth.serverAddr = backendIO.SourceAddr().String()
 	auth.clientAddr = clientIO.SourceAddr().String()
 
 	proxyCapability := auth.supportedServerCapabilities
--- a/pkg/proxy/backend/authenticator.go
+++ b/pkg/proxy/backend/authenticator.go
@@ -15,13 +15,17 @@
 package backend
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/binary"
 	"fmt"
 	"net"
 	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/pingcap/TiProxy/lib/util/errors"
+	"github.com/pingcap/TiProxy/pkg/manager/router"
 	pnet "github.com/pingcap/TiProxy/pkg/proxy/net"
 	"github.com/pingcap/tidb/parser/mysql"
 	"github.com/pingcap/tidb/util/hack"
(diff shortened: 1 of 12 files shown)


pkg/proxy/backend/authenticator.go
```diff
@@ -101,8 +105,9 @@ func (auth *Authenticator) verifyBackendCaps(logger *zap.Logger, backendCapabili
 	return nil
 }
 
-func (auth *Authenticator) handshakeFirstTime(logger *zap.Logger, clientIO *pnet.PacketIO, handshakeHandler HandshakeHandler,
-	getBackend backendIOGetter, frontendTLSConfig, backendTLSConfig *tls.Config) error {
+func (mgr *BackendConnManager) handshakeFirstTime(logger *zap.Logger, clientIO, backendIO *pnet.PacketIO, handshakeHandler HandshakeHandler, frontendTLSConfig, backendTLSConfig *tls.Config) error {
```
djshow832, line 108: I don't like this change.
The code is all about authentication but you moved it to `BackendConnManager`.
Besides, it's also not a good idea to introduce the concept `router` in the authentication.

xhebox: Considering that they are actually coded closely despite the concepts are separated and clear, I prefer integrating them together.

Here are some arguments:

1. `authenticator` actually holds the customized context of `handshakeHandler` anyway.
2. It contains proxy protocol. And it can only be done in the authenticator.
3. `handshakeHandler` is passed to customize `HandshakeResp`. To archive the full logic of serverless gateway, maybe there will be more customization.
4. Actually `BackendConnMgr` depends on internal fields of authenticator, e.g. capability, which should be connection states instead of authenticator-only information, IMO.

But I agree that maybe this function is too long. One solution would be moving the `getBackendIO` into a separate function, then it still looks like the old code. The only difference would be the function signature.

xhebox: I would say that the problem is in MYSQL protocol. The handshake process itself is not all about authentication, but also load balance, server capability, session tracking, TLS security, compression, etc... Those are not part of the authenticator. Instead, authentication is part of the handshake.

djshow832: You can rename `Authenticator` to `Handshaker` or whatever that is related to handshake, but it processes all the things about handshaking. `BackendConnManager` splits some logic to `Authenticator` and `CmdProcessor` to make itself simpler.
If you weaken the boundary of these logics, it again makes them coupled and even into one struct finally.

djshow832: If it's all about readability, you can pass a `HandshakeHandlerWrapper` to `Authenticator`, which adds a `GetBackendIO()` function to `HandshakeHandler`.

xhebox: I really don't mind it is one struct or two. There are lots of mega structs in various projects. I can accept something like:

```
type CommandProcessor struct {
   ...
}

type Conn struct {
  // common fields
  ...
  // authentication only fields
  ...
  // command processor fields
  cmdprc *CommandProcessor
}
```

I cares more about if the definition of functions are split accordingly. That is no coupling for me.

xhebox: What about now? Now `handshakeFirstTime` is still a method of `BackendConnMgr` located at `authenticator.go` and invoke `mgr.getBackendIO()` to do the routing. The only difference is the function signature.

djshow832 (approved):

Issue #161, opened 2022-12-26
Support dedicated tier


djshow832:
## Development Task

Some modifications to adapt to the dedicated tier:
- #159
- #164
- Retry connecting backends when the backends are not available
- #165
- #167 
- #169
- pingcap/tidb-operator#4821

PR #159, opened 2022-12-23
server, backend: graceful shutdown

djshow832:
### What problem does this PR solve?

Issue Number: close #xxx

Problem Summary:

What is changed and how it works:

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Notable changes

- [ ] Has configuration change
- [ ] Has HTTP API interfaces change (Don't forget to [add the declarative for API](https://github.com/tikv/pd/blob/master/docs/development.md#updating-api-documentation))
- [ ] Has tiproxyctl change
- [ ] Other user behavior changes

### Release note

Please refer to [Release Notes Language Style Guide](https://pingcap.github.io/tidb-dev-guide/contribute-to-tidb/release-notes-style-guide.html) to write a quality release note.

```release-note
None
```

PR #164, opened 2022-12-27
backend: fix connection won't close if graceful shutdown starts during handshake

djshow832:

Issue Number: ref #161

Problem Summary:

- In #159, if graceful shutdown starts during or before handshaking, the `statusConnected` or `statusHandshaked` overrides `statusNotifyClose`, which makes `statusNotifyClose` not work.
- When TiProxy receives a `SIGINT` signal, the `cmd.Context()` is done, and it will stop some components, such as the goroutine which processes `signalGracefulClose`. Thus, the `BackendConnMgr` won't shutdown immediately.

- Rename `connStatus` to `closeStatus` and it only manages the status of closing.
- Create another context in `SQLServer`. I've checked the code of the gateway, it won't affect the gateway.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

- graceful-shutdown=10 and stop the TiProxy
- graceful-shutdown=10, send a `begin` statement and stop the TiProxy
- graceful-shutdown=10, send a `begin` statement, stop the TiProxy, and send a `commit` statement
- graceful-shutdown=0, send a `begin` statement and stop the TiProxy

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


conf/proxy.yaml
```diff
@@ -4,7 +4,7 @@ proxy:
   tcp-keep-alive: true
   max-connections: 1000
   pd-addrs: "127.0.0.1:2379"
-  # require-backend-tls: true
+  require-backend-tls: false
```
xhebox: You can change code https://github.com/pingcap/TiProxy/blob/main/lib/config/proxy.go#L119 instead.

pkg/proxy/client/client_conn.go
```diff
@@ -72,8 +72,8 @@ func (cc *ClientConnection) Run(ctx context.Context) {
 	}
 
 clean:
-	// graceful close
-	if errors.Is(err, os.ErrDeadlineExceeded) {
+	// graceful close or force close when shutting down
+	if errors.Is(err, os.ErrDeadlineExceeded) || errors.Is(err, net.ErrClosed) {
```
xhebox, line 76: I think we can ignore errors with `!clientErr && (errors.Is(err, os.ErrDeadlineExceeded) || errors.Is(err, net.ErrClosed))` for logging, instead of direct return.

xhebox (commented):
Rest LGTM

conf/proxy.yaml
djshow832: Weird, I reverted it before you reviewed it...

pkg/proxy/client/client_conn.go
djshow832: Do you mean still log them? They will be too many.
BTW, clientErr is true for these cases.

xhebox: > Do you mean still log them? They will be too many.

No I meant to ignore them. But we should judge `clientErr`, too, since it may come from server somehow, too.

>  BTW, clientErr is true for these cases.

Oh, yes, I wrote it wrong.  Maybe:

```
if clientErr &&
 (errors.Is(err, io.EOF) || errors.Is(err, os.ErrDeadlineExceeded) || errors.Is(err, net.ErrClosed)) {
   return
}
cc.logger.Info(...)
```

xhebox (approved):

Issue #168, opened 2023-01-03
Retry different backends when establishing connections


djshow832:
## Development Task

In the serverless tier, the backends may be recycled when establishing connections, so TiProxy needs to try other backends if it fails.
In the dedicated tier, the backends may be both marked as unhealthy but one of them may be healthy, so TiProxy needs to try unhealthy ones if there's no choice.

PR #171, opened 2023-01-04
router, backend: retry connecting to different backends

djshow832:

Issue Number: ref #168

Problem Summary:
To adapt to the serverless tier, retry connecting to different backends one by one.
Note that this PR has not adapted to the dedicated tier yet.

- `Router.Route()` returns `BackendSelector` now, which returns a backend each time for calling `Next()`.
- `BackendConnMgr` tries connecting to the backends at max 5 seconds.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -160,11 +160,12 @@ func (mgr *BackendConnManager) getBackendIO(ctx ConnContext, auth *Authenticator
...
 	bctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	selector := r.Route()
 	addr, err := backoff.RetryNotifyWithData(
 		func() (string, error) {
-			addr, err := r.Route(mgr)
-			if !errors.Is(err, router.ErrNoInstanceToSelect) {
-				return addr, backoff.Permanent(err)
+			addr := selector.Next()
+			if len(addr) == 0 {
+				return "", router.ErrNoInstanceToSelect
```
xhebox, line 168: Just a note, `backoff.Permanent` will stop retrying/backoffing. Normal errors won't stop the loop.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -178,15 +179,31 @@ func (mgr *BackendConnManager) getBackendIO(ctx ConnContext, auth *Authenticator
...
+					auth.serverAddr = addr
+					return mgr.backendConn.PacketIO(), nil
+				}
+				// Bad luck: the backend has been recycled or shut down just after the selector returns it.
+				if ignoredErr := backendConn.Close(); ignoredErr != nil {
+					mgr.logger.Error("close backend connection failed", zap.String("addr", addr), zap.Error(ignoredErr))
+				}
+			}
+			addr = selector.Next()
+		}
```
xhebox, line 202: I think we can integrate this loop with `backoff.RetryNotifyWithData` loop above. It is basically same thing with and without `Connect()`.

I think the following will be good:

```
for retries := 0; retries < maxRetries; retries++ {
   selector.Reset()
   backoff.backoff.RetryNotifyWithData(
		func() (*pnet.PacketIO, error) {
   ...
}
```

Or maybe the following, then backoff is more controlled by time instead of retry times:

```
selector := r.Route()
backoff.backoff.RetryNotifyWithData(
		func() (*pnet.PacketIO, error) {
                    addr := selector.Next()
                    if len(addr) == 0 { selector.Reset(); addr =  selector.Next() }
```

pkg/manager/router/backend_selector.go
```diff
@@ -0,0 +1,38 @@
...
+package router
+
+type BackendSelector struct {
+	excluded  []string
+	cur       string
+	routeOnce func(excluded []string) string
+	addConn   func(addr string, conn RedirectableConn) error
+}
+
+func (bs *BackendSelector) Reset() {
```
xhebox, line 24: I think we can just delete `Reset()` API, making selector always returning a possible addr until we timeout, or we call `Succeed`.

pkg/proxy/backend/backend_conn_mgr_test.go
```diff
@@ -683,3 +686,42 @@ func TestGracefulCloseBeforeHandshake(t *testing.T) {
...
+		onHandshake: func(connContext ConnContext, s string, err error) {
+			if err != nil && len(s) > 0 {
+				badAddrs[s] = struct{}{}
+			}
+		},
+	}
+	mgr := NewBackendConnManager(logger.CreateLoggerForTest(t), handler, 0, false, false)
+	auth := &Authenticator{}
+	for i := 0; i <= len(listeners); i++ {
+		io, err := mgr.getBackendIO(auth, auth, nil)
```
xhebox, line 714: Can't we just use `mgr.authenticator`?

pkg/proxy/backend/backend_conn_mgr_test.go
```diff
@@ -683,3 +686,42 @@ func TestGracefulCloseBeforeHandshake(t *testing.T) {
...
+		if i < len(listeners) {
+			require.NoError(t, err)
+			err = io.Close()
+			require.NoError(t, err)
+			err = listeners[i].Close()
+			require.NoError(t, err)
```
xhebox, line 720: Maybe use `t.Cleanup` to close listeners.

pkg/proxy/backend/backend_conn_mgr_test.go
```diff
@@ -683,3 +686,42 @@ func TestGracefulCloseBeforeHandshake(t *testing.T) {
...
+		},
+	}
+	mgr := NewBackendConnManager(logger.CreateLoggerForTest(t), handler, 0, false, false)
+	auth := &Authenticator{}
+	for i := 0; i <= len(listeners); i++ {
+		io, err := mgr.getBackendIO(auth, auth, nil)
+		if i < len(listeners) {
+			require.NoError(t, err)
+			err = io.Close()
+			require.NoError(t, err)
```
xhebox, line 718: I think the following will be more robust:

```
if err = nil {
  require.NoError(t, io.Close()
}
if i < len(listeners)
```

xhebox (commented):
Rest LGTM

pkg/proxy/backend/backend_conn_mgr.go
djshow832: I know. The only error it may return is no instances to route, and it should continue to try.

pkg/proxy/backend/backend_conn_mgr_test.go
djshow832: It's not just cleanup, it's testing that the selector still returns an available listener even if some are closed.

pkg/manager/router/backend_selector.go
djshow832: I'm afraid it will be not easy to control backoff. Say, I want to iterate all the backends first, and then sleep for 100ms before continuing the next loop, and then sleep for 200ms before the next.

xhebox: Makes sense.

pkg/proxy/backend/authenticator.go
```diff
@@ -160,7 +161,7 @@ func (auth *Authenticator) handshakeFirstTime(logger *zap.Logger, clientIO *pnet
 	auth.attrs = resp.Attrs
 
 	// In case of testing, backendIO is passed manually that we don't want to bother with the routing logic.
-	backendIO, err := getBackendIO(auth, auth, resp)
+	backendIO, err := getBackendIO(auth, auth, resp, 5*time.Second)
```
xhebox, line 164: There is a constant, `DialTimeout` in `backend_conn.go`.

xhebox (approved):

pkg/proxy/backend/authenticator.go
djshow832: It's different. The one passed to `getBackendIO` is the total timeout for all retries.
BTW, `DialTimeout` is also 5 seconds. Maybe it should be less?

PR #176, opened 2023-01-06
api: refactor config API to adapt tidb-operator/TOML changes

xhebox:

Issue Number: ref #161 

Problem Summary: Due to we have adapt `ConfigWrapper` in `tidb-operator`, it becomes harder to set config for each component separately. Further more, we need to issue two requests for each instace. Instead, we just upload the whole config, and pick updatable part now.

1.  `config [log | proxy] [get | put]` is replaced by one single command `tiproxyctl config get/put`
2. Watch channels for `XXXOnlineConfig` are replaced by `chan *config.Config`
3. Init values are inited in `ConfigManager.Init()`
4. Stablize log config by removing file size check

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [x] Has tiproxyctl change

### Release note

```release-note
None
```


pkg/manager/logger/manager_test.go
```diff
@@ -78,10 +80,7 @@ func TestUpdateCfg(t *testing.T) {
 				}
 			},
 			check: func(files []os.FileInfo) bool {
-				if len(files) != 1 {
-					return false
-				}
-				return files[0].Size() >= int64(2500*1024)
```
xhebox, line 84: Rotation logger is not going to flush all data immediately into files, making file size always a little bit less than the expected value. Just remove it.

djshow832: But I saw you have added `log.Sync()` to flush it.
Besides, each case waits for at most 3 seconds.

xhebox: Yeah, there is `log.Sync()`. But it is stub, `lumberjack` does not have a `sync()`. Check https://github.com/pingcap/TiProxy/blob/main/lib/util/cmd/syncer.go#L42-L48 .

Considering it is always failing and `lumberjack` is actually a well-tested library. I think it is just fine not to test the size behavior.

pkg/manager/logger/manager.go
```diff
@@ -51,20 +51,23 @@ func NewLoggerManager(cfg *config.Log) (*LoggerManager, *zap.Logger, error) {
...
 
-func (lm *LoggerManager) watchCfg(ctx context.Context, cfgCh <-chan *config.LogOnline) {
+func (lm *LoggerManager) watchCfg(ctx context.Context, cfgch <-chan *config.Config) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case cfg := <-cfgCh:
+		case acfg := <-cfgch:
+			cfg := &acfg.Log.LogOnline
```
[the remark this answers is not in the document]
xhebox: No delete in the current implementation. And it is ensured to be valid pointer:

```
	var c config.Config
		if k.Type == KVEventDel {
			ch <- &c
		} else {
```

pkg/manager/logger/manager_test.go
djshow832: It never failed before. Does it fail in this PR?

xhebox: Yes, very stable failure on my machine. I guess it is still not stable.

djshow832: Does it fail on your machine before this PR?

xhebox: No, but I can't figure out any difference. I've debugged that `config options` are updated correctly. It is just failing with the correct test case.

djshow832: I'll try it on my machine. I'm afraid there's some other bug.

xhebox: BTW, it failed on CI before.

djshow832 (approved):

Issue #180, opened 2023-01-10
Send the error to the client when the handler encounters an error


djshow832:
## Development Task

Currently, when the serverless tier encounters an error, the error is logged but not returned to the client. The client cannot figure out the error easily.
We need to wrap the error into an MySQL error and return it to the client.

PR #182, opened 2023-01-10
router, backend: stop retrying connecting to backends if the fetcher fails

### LINKED ISSUE #179: Stop retrying connecting to backends if the fetcher fails
## Development Task

Currently, when the backend fetcher returns an empty backend list, the proxy keeps retrying for 5 seconds.
However, the serverless tier may encounter some problems that can't be recovered immediately. In this case, the fetcher can return an error, indicating that the proxy doesn't need to retry.

djshow832:

Issue Number: close #179

Problem Summary:
Currently, when the backend fetcher returns an empty backend list, the proxy keeps retrying for 5 seconds.
However, the serverless tier may encounter some problems that can't be recovered immediately. In this case, the fetcher can return an error, indicating that the proxy doesn't need to retry.

- Add a return value of `error` type for `BackendFetcher.GetBackendList()`
- Add the return value for `Observer`, `BackendEventReceiver`, `BackendSelector`, and so on
- Stop retrying connecting if the error is not nil

Note that https://github.com/pingcap/TiProxy/issues/180 will be fixed in another PR.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/manager/router/backend_observer.go
+++ b/pkg/manager/router/backend_observer.go
@@ -185,6 +185,7 @@ func (bo *BackendObserver) observe(ctx context.Context) {
 		backendInfo, err := bo.fetcher.GetBackendList(ctx)
 		if err == nil {
 			if lastErr != nil {
+				bo.logger.Info("fetching backends succeeds now")
 				bo.eventReceiver.OnObserveError(err)
 			}
 			backendStatus := bo.checkHealth(ctx, backendInfo)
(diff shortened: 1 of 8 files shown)


pkg/manager/router/backend_observer.go
```diff
@@ -178,13 +180,26 @@ func (bo *BackendObserver) Refresh() {
...
 	for ctx.Err() == nil {
-		backendInfo := bo.fetcher.GetBackendList(ctx)
-		backendStatus := bo.checkHealth(ctx, backendInfo)
-		if ctx.Err() != nil {
-			return
+		backendInfo, err := bo.fetcher.GetBackendList(ctx)
+		if err == nil {
+			if lastErr != nil {
+				bo.logger.Info("fetching backends succeeds now")
+				bo.eventReceiver.OnObserveError(err)
```
xhebox, line 189: How about use only callback `OnObserve(map[string]XXX, error)` or `OnObserve(ObserveEvent)`?
I feel it is a little bit complicated that `backendChangeEvent` and `backendChangeErrorEvent` are parallel.

It seems more reasonable to invoke `OnXXXEvent` everytime:
1. polled the fetcher, and something is returned
2. gateway asynchronizely pushed some events to the observer

Issue #229, opened 2023-03-03
Queries may block when the backend is down


djshow832:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

Start queries and stop the TiDB pod

### 2. What did you expect to see? (Required)
Some connections are closed

### 3. What did you see instead (Required)
Some connections wait for response for 20 minutes

### 4. What is your version? (Required)

master

PR #231, opened 2023-03-05
proxy: provide keepalive helpers

xhebox:

Issue Number: ref #229

Problem Summary: Cross-platform keepalive support using SyscallConn located in /pkg/proxy/keepalive
### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/lib/config/proxy.go
+++ b/lib/config/proxy.go
@@ -45,10 +45,10 @@ type Metrics struct {
 }
 
 type KeepAlive struct {
-	Disabled bool          `yaml:"disabled,omitempty" toml:"disabled,omitempty" json:"disabled,omitempty"`
-	Cnt      int           `yaml:"cnt,omitempty" toml:"cnt,omitempty" json:"cnt,omitempty"`
-	Idle     time.Duration `yaml:"idle,omitempty" toml:"idle,omitempty" json:"idle,omitempty"`
-	Intvl    time.Duration `yaml:"intvl,omitempty" toml:"intvl,omitempty" json:"intvl,omitempty"`
+	Enabled bool          `yaml:"enabled,omitempty" toml:"enabled,omitempty" json:"enabled,omitempty"`
+	Cnt     int           `yaml:"cnt,omitempty" toml:"cnt,omitempty" json:"cnt,omitempty"`
+	Idle    time.Duration `yaml:"idle,omitempty" toml:"idle,omitempty" json:"idle,omitempty"`
+	Intvl   time.Duration `yaml:"intvl,omitempty" toml:"intvl,omitempty" json:"intvl,omitempty"`
 }
 
 type ProxyServerOnline struct {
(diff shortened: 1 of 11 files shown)


lib/config/proxy.go
```diff
@@ -114,7 +123,7 @@ func NewConfig() *Config {
 	var cfg Config
 
 	cfg.Proxy.Addr = "0.0.0.0:6000"
-	cfg.Proxy.TCPKeepAlive = true
+	cfg.Proxy.FrontendKeepalive.Cnt = 1
```
djshow832, line 126: Set the frontend count to 1? I think the default value is better. That is, 0 means not to update it.

pkg/proxy/keepalive/keepalive.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/pingcap/TiProxy/lib/config"
+	"github.com/pingcap/TiProxy/lib/util/errors"
+)
+
+var (
+	ErrKeepAlive = errors.New("failed to set keepalive")
+)
+
+func SetKeepalive(conn net.Conn, cfg config.KeepAlive) error {
+	if !cfg.Enabled {
```
djshow832, line 29: It should be enabled by default; if it's false, call setKeepalive(false).

pkg/proxy/keepalive/keepalive_darwin.go
```diff
@@ -0,0 +1,52 @@
...
+	return errors.Collect(ErrKeepAlive, serr, syscn.Control(func(fd uintptr) {
+		serr = syscall.SetsockoptInt(int(fd), syscall.SOL_SOCKET, syscall.SO_KEEPALIVE, 1)
+		if serr != nil {
+			return
+		}
+		serr = syscall.SetsockoptInt(int(fd), syscall.IPPROTO_TCP, syscall.TCP_KEEPALIVE, int(cfg.Idle.Seconds()))
+		if serr != nil {
+			return
+		}
+		serr = syscall.SetsockoptInt(int(fd), syscall.IPPROTO_TCP, _TCP_KEEPCNT, cfg.Cnt)
```
djshow832, line 43: What will happen if cnt==0? It's 0 by default.
I think if cnt==0/idle==0/intvl==0, do not set it. Or set a default value to them.

pkg/proxy/keepalive/keepalive.go
djshow832: Is it possible that it's true by default? So if it's false, you need to disable it explicitly.

xhebox: What about now?

djshow832: - It's a bad idea to name a configuration item `disabled`, because it's inconsistent with other configurations. Previous TiDB users complained about this.
- It doesn't solve the problem: the keep alive may be enabled and then disabled multiple times, since the configuration is supposed to be dynamic. So you need to set it explicitly every time it changes.

xhebox: There is no way that it can be enabled by default. There are total 4 syscall options. I split them into `SetKeepalive` and `SetKeepaliveParams`. See if that suits you.

djshow832 (approved):

PR #242, opened 2023-03-08
config: adjust keep alive parameters and add comments

djshow832:

Issue Number: close none

Problem Summary:
Now that we have TCP timeout, we can make keep alive parameters less aggressive, because it only works when the client is idle.
Besides, 30s is too long for a backend to respond. The client cannot wait for 30s.

- Adjust healthy keep alive to 75s and timeout to 15s
- Adjust unhealthy keep alive to 15s and timeout to 5s

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/lib/config/proxy.go
+++ b/lib/config/proxy.go
@@ -47,12 +47,12 @@ type Metrics struct {
 type KeepAlive struct {
 	Enabled bool `yaml:"enabled,omitempty" toml:"enabled,omitempty" json:"enabled,omitempty"`
 	// Idle, Cnt, and Intvl works only when the connection is idle. User packets will interrupt keep-alive.
-	// If the peer crashes and doesn't send FIN, the connection will be closed within Idle+Cnt*Intvl.
+	// If the peer crashes and doesn't send any packets, the connection will be closed within Idle+Cnt*Intvl.
 	Idle  time.Duration `yaml:"idle,omitempty" toml:"idle,omitempty" json:"idle,omitempty"`
 	Cnt   int           `yaml:"cnt,omitempty" toml:"cnt,omitempty" json:"cnt,omitempty"`
 	Intvl time.Duration `yaml:"intvl,omitempty" toml:"intvl,omitempty" json:"intvl,omitempty"`
 	// Timeout is the timeout of waiting ACK. It works for both user packets and keep-alive.
-	// For keep-alive, the timeout of waiting ACK is min(Timeout, Cnt*Intvl).
+	// It is suggested to be equal or close to Cnt*Intvl.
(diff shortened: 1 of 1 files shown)


lib/config/proxy.go
```diff
@@ -45,17 +45,25 @@ type Metrics struct {
 }
 
 type KeepAlive struct {
-	Enabled bool          `yaml:"enabled,omitempty" toml:"enabled,omitempty" json:"enabled,omitempty"`
-	Idle    time.Duration `yaml:"idle,omitempty" toml:"idle,omitempty" json:"idle,omitempty"`
-	Cnt     int           `yaml:"cnt,omitempty" toml:"cnt,omitempty" json:"cnt,omitempty"`
-	Intvl   time.Duration `yaml:"intvl,omitempty" toml:"intvl,omitempty" json:"intvl,omitempty"`
+	Enabled bool `yaml:"enabled,omitempty" toml:"enabled,omitempty" json:"enabled,omitempty"`
+	// Idle, Cnt, and Intvl works only when the connection is idle. User packets will interrupt keep-alive.
+	// If the peer crashes and doesn't send FIN, the connection will be closed within Idle+Cnt*Intvl.
```
xhebox, line 50: It is actually waiting for `ACK`.

lib/config/proxy.go
```diff
@@ -45,17 +45,25 @@ type Metrics struct {
...
+	Idle  time.Duration `yaml:"idle,omitempty" toml:"idle,omitempty" json:"idle,omitempty"`
+	Cnt   int           `yaml:"cnt,omitempty" toml:"cnt,omitempty" json:"cnt,omitempty"`
+	Intvl time.Duration `yaml:"intvl,omitempty" toml:"intvl,omitempty" json:"intvl,omitempty"`
+	// Timeout is the timeout of waiting ACK. It works for both user packets and keep-alive.
+	// For keep-alive, the timeout of waiting ACK is min(Timeout, Cnt*Intvl).
```
xhebox, line 55: Not exactly, if `timeout > cnt*intvl`, actually linux will send more probes.

So the best parameter will be equal or very close. A large timeout value is not suitable, too.

lib/config/proxy.go
djshow832: I mean, when it's idle. The peer should send FIN before it closes. But if it crashes, it doesn't send FIN.

xhebox: I knew. But if it is not crashed, it will also be disconnected. For example, delayed ack feature.

xhebox (approved):

Issue #233, opened 2023-03-05
Replace other atomic interfaces to sync/atomic


djshow832:
## Development Task

We have some different atomic interfaces:
- go.uber.org/atomic
- unsafe.Pointer
- sync/atomic

We can replace another 2 with sync/atomic to be consistent.

PR #271, opened 2023-04-20
backend: use atomic.Pointer for *signalRedirect

xhebox:

Issue Number: ref #233

Problem Summary: Cleanup.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/manager/router/router_score.go
+++ b/pkg/manager/router/router_score.go
@@ -373,18 +373,16 @@ func (router *ScoreBasedRouter) rebalance(maxNum int) {
 			break
 		}
 		conn := ce.Value
-		if conn.IsRedirectable() {
-			router.logger.Info("begin redirect connection", zap.Uint64("connID", conn.ConnectionID()),
-				zap.String("from", busiestBackend.addr), zap.String("to", idlestBackend.addr),
-				zap.Int("from_score", busiestBackend.score()), zap.Int("to_score", idlestBackend.score()))
-			busiestBackend.connScore--
-			router.adjustBackendList(busiestEle)
-			idlestBackend.connScore++
-			router.adjustBackendList(idlestEle)
-			conn.phase = phaseRedirectNotify
-			conn.lastRedirect = curTime
-			conn.Redirect(idlestBackend.addr)
-		}
+		router.logger.Info("begin redirect connection", zap.Uint64("connID", conn.ConnectionID()),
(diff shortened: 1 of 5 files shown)


pkg/manager/router/router_score.go
```diff
@@ -365,16 +366,18 @@ func (router *ScoreBasedRouter) rebalance(maxNum int) {
...
-			zap.String("from", busiestBackend.addr), zap.String("to", idlestBackend.addr),
-			zap.Int("from_score", busiestBackend.score()), zap.Int("to_score", idlestBackend.score()))
-		busiestBackend.connScore--
-		router.adjustBackendList(busiestEle)
-		idlestBackend.connScore++
-		router.adjustBackendList(idlestEle)
-		conn.phase = phaseRedirectNotify
-		conn.lastRedirect = curTime
-		conn.Redirect(idlestBackend.addr)
+		if conn.IsRedirectable() {
```
djshow832, line 369: It will cause a deadloop: the next time, the connection is chosen again, and again, and again.
I don't think it's very necessary to add this check: `mgr.redirectInfo.Load() == nil` is ensured and closing status is handled.

xhebox: > It will cause a deadloop: the next time, the connection is chosen again, and again, and again.

How? You already filtered it by `conn.phase`.

> I don't think it's very necessary to add this check: mgr.redirectInfo.Load() == nil is ensured and closing status is handled.

True. But I want to leverage these checks into `conn`, instead of tracking them in the router.

djshow832: If you check the `phase` or `redirectInfo` only in `conn`, then it will cause deadloop. If you also check it in the router, you are still tracking it in the router. I don't get your purpose.

xhebox: > If you check the phase or redirectInfo only in conn

I don't get why there will be deadloop.

> If you also check it in the router, you are still tracking it in the router. I don't get your purpose.

I want to check `phase/redirectInfo` in the router by methods of `conn`. Basically I want to move the functionality of `connWrapper` into `conn`, reducing the complexity of router.

It matters when I want to separate router/conn by interfaces/channels, or even RPC.

djshow832 (approved):

PR #379, opened 2023-10-12
[WIP] handle prefix mismatch 

AmoebaProtozoa:

Sometimes, due to router cache out of date, an authentication request will be directed to a unexpected tidb that belongs to another keyspace and rejected. A special error code (20003) is added in this case to differentiate this from regular authentication error. 

This error should then be treated not as MySQL error but as proxy error

### DIFF
--- a/pkg/proxy/backend/cmd_processor.go
+++ b/pkg/proxy/backend/cmd_processor.go
@@ -17,6 +17,11 @@ const (
 	StatusQuit
 	StatusPrepareWaitExecute
 	StatusPrepareWaitFetch
+
+	// errUserPrefixMismatch is a special error that indicates the request is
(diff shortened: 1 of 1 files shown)


better0332 (approved):

djshow832:
What's the purpose of treating it as a proxy error? Returning a different error message to the client? Or setting a different `ErrorSource` in the `ConnContext`?

djshow832:
TiProxy is supposed to be generally used, so please move the customized logic into gateway.
E.g. Add a function for judging whether it's a proxy error in `HandshakeHandler`, say `HandshakeHandler.ErrorSource(error) ErrorSource`, and the Gateway customizes the implementation.

AmoebaProtozoa:
closing PR, will handle this at gateway level instead

PR #568, opened 2024-06-20
router, backend: pause session when last backend become unhealthy

clark1013:

Problem Summary: 
- Introduce a new feature that TiProxy can pause session.

-  TiProxy store session states when last backend become unhealthy. 
- When the session is saved, tiproxy won't closed client connection when backend connection is closed.
- The session can be restored when client become active again to a new backend.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/manager/router/backend_observer.go
+++ b/pkg/manager/router/backend_observer.go
@@ -144,7 +144,6 @@ func (bo *BackendObserver) observe(ctx context.Context) {
 		startTime := monotime.Now()
 		backendInfo, err := bo.fetcher.GetBackendList(ctx, refresh)
 		refresh = false
-		// bo.logger.Info("debug GetBackendList", zap.Any("backendInfo", backendInfo))
 		if err != nil {
 			bo.logger.Error("fetching backends encounters error", zap.Error(err))
 			bo.eventReceiver.OnBackendChanged(nil, err)
--- a/pkg/manager/router/backend_observer.go
+++ b/pkg/manager/router/backend_observer.go
@@ -139,9 +139,12 @@ func (bo *BackendObserver) Refresh() {
 }
 
 func (bo *BackendObserver) observe(ctx context.Context) {
+	refresh := false
 	for ctx.Err() == nil {
 		startTime := monotime.Now()
-		backendInfo, err := bo.fetcher.GetBackendList(ctx)
+		backendInfo, err := bo.fetcher.GetBackendList(ctx, refresh)
+		refresh = false
+		// bo.logger.Info("debug GetBackendList", zap.Any("backendInfo", backendInfo))
 		if err != nil {
 			bo.logger.Error("fetching backends encounters error", zap.Error(err))
 			bo.eventReceiver.OnBackendChanged(nil, err)
@@ -160,6 +163,7 @@ func (bo *BackendObserver) observe(ctx context.Context) {
 			select {
 			case <-time.After(wait):
 			case <-bo.refreshChan:
+				refresh = true
 			case <-ctx.Done():
 				return
 			}
--- a/pkg/manager/router/backend_observer.go
+++ b/pkg/manager/router/backend_observer.go
@@ -139,6 +139,8 @@ func (bo *BackendObserver) Refresh() {
 }
 
 func (bo *BackendObserver) observe(ctx context.Context) {
+	// refresh means the fetcher request is triggerd by calling Refresh() actively.
+	// It typically indicates their are no available backends.
 	refresh := false
 	for ctx.Err() == nil {
 		startTime := monotime.Now()
--- a/pkg/manager/router/router_score.go
+++ b/pkg/manager/router/router_score.go
@@ -33,14 +33,17 @@ type ScoreBasedRouter struct {
 	backends     *glist.List[*backendWrapper]
 	observeError error
 	// Only store the version of a random backend, so the client may see a wrong version when backends are upgrading.
-	serverVersion string
+	serverVersion     string
+	enableZeroBackend bool
+	inZeroBackendMode bool
 }
 
 // NewScoreBasedRouter creates a ScoreBasedRouter.
-func NewScoreBasedRouter(logger *zap.Logger) *ScoreBasedRouter {
+func NewScoreBasedRouter(logger *zap.Logger, enableZeroBackend bool) *ScoreBasedRouter {
 	return &ScoreBasedRouter{
-		logger:   logger,
-		backends: glist.New[*backendWrapper](),
+		logger:            logger,
+		backends:          glist.New[*backendWrapper](),
+		enableZeroBackend: enableZeroBackend,
 	}
 }
 
@@ -314,9 +317,39 @@ func (router *ScoreBasedRouter) OnBackendChanged(backends map[string]*BackendHea
 			backend.setHealth(*health)
 			router.adjustBackendList(be, true)
 		}
+		if health.Status == StatusHealthy {
+			router.inZeroBackendMode = false
+		}
 	}
+	// router.logger.Info(
+	// 	"debug backend change",
+	// 	zap.Int("len", router.backends.Len()),
+	// 	zap.Bool("zero_backend", router.inZeroBackendMode),
+	// 	zap.Any("backends", router.backends),
+	// )
 	if len(backends) > 0 {
 		router.updateServerVersion()
+	}
+	if router.enableZeroBackend && len(backends) > 0 {
+		inZeroBackendMode := true
+		for be := router.backends.Front(); be != nil; be = be.Next() {
+			backend := be.Value
+			if backend.Healthy() {
+				inZeroBackendMode = false
+				break
+			}
+		}
+		router.inZeroBackendMode = inZeroBackendMode
+		if inZeroBackendMode {
+			router.logger.Info("last backend become unhealthy, notify connections to save session")
+			for be := router.backends.Front(); be != nil; be = be.Next() {
+				backend := be.Value
+				for ele := backend.connList.Front(); ele != nil; ele = ele.Next() {
+					conn := ele.Value
+					conn.SaveSession()
+				}
+			}
+		}
 	}
 }
 
@@ -430,6 +463,10 @@ func (router *ScoreBasedRouter) ServerVersion() string {
 	version := router.serverVersion
 	router.Unlock()
 	return version
+}
+
+func (router *ScoreBasedRouter) InZeroBackendMode() bool {
+	return router.inZeroBackendMode
 }
 
 // Close implements Router.Close interface.
--- a/pkg/manager/router/router_score.go
+++ b/pkg/manager/router/router_score.go
@@ -21,6 +21,24 @@ const (
 
 var _ Router = &ScoreBasedRouter{}
 
+type RouterConfig struct {
+	EnablePause bool
+}
+
+func NewDefaultRouterConfig() *RouterConfig {
+	return &RouterConfig{
+		EnablePause: false,
+	}
+}
+
+type RouterConfigFunc func(*RouterConfig)
+
+func WithPauseEnabled() RouterConfigFunc {
+	return func(cfg *RouterConfig) {
+		cfg.EnablePause = true
+	}
+}
+
 // ScoreBasedRouter is an implementation of Router interface.
 // It routes a connection based on score.
 type ScoreBasedRouter struct {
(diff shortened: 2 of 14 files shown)


pkg/manager/router/router_score.go
```diff
@@ -318,6 +375,32 @@ func (router *ScoreBasedRouter) OnBackendChanged(backends map[string]*BackendHea
...
+			if backend.Healthy() {
+				pause = false
+				break
+			}
+		}
+		if pause {
+			router.logger.Info("last backend become unhealthy, notify connections to pause")
+			for be := router.backends.Front(); be != nil; be = be.Next() {
+				backend := be.Value
+				backend.connScore--
```
djshow832, line 391: `connScore` should be updated when pausing each connection.

pkg/manager/router/router_score.go
```diff
@@ -318,6 +375,32 @@ func (router *ScoreBasedRouter) OnBackendChanged(backends map[string]*BackendHea
...
+				for ele := backend.connList.Front(); ele != nil; ele = ele.Next() {
+					conn := ele.Value
+					switch conn.phase {
+					case phasePauseNotify, phaseRedirectNotify:
+						continue
```
djshow832, line 396: If a connection is redirecting and doesn't pause, you need to handle it. E.g. notify it later.

pkg/manager/router/router_score.go
```diff
@@ -269,24 +295,55 @@ func (router *ScoreBasedRouter) onRedirectFinished(from, to string, conn Redirec
...
+	router.Lock()
+	defer router.Unlock()
+	be := router.ensureBackend(addr, true)
+	connWrapper := router.getConnWrapper(conn).Value
+	if succeed {
+		router.removeConn(be, router.getConnWrapper(conn))
+		router.addPausedConn(connWrapper)
+		connWrapper.phase = phaseNone
+	} else {
+		connWrapper.phase = phasePauseFail
```
djshow832, line 320: You need to increase connScore back.

pkg/manager/router/router_score.go
```diff
@@ -408,6 +493,7 @@ func (router *ScoreBasedRouter) ConnCount() int {
 		backend := be.Value
 		j += backend.connList.Len()
 	}
+	j += router.pausedConnList.Len()
```
djshow832, line 496: If `enablePause=false`, `pausedConnList` is nil. You can just always create `pausedConnList`.

pkg/manager/router/router_score.go
clark1013: In serverless scenario. If conn is redirecting while pausing, it indicates the backend become alive while pausing. I think it's OK to just let it migrate to the new backend. Can I just leave a log here? And maybe we can handle it better   when pause be triggered by other events in the future.

pkg/manager/router/router_score.go
clark1013: OK, I will update it.

pkg/manager/router/router_score.go
djshow832: One question, if a connection is in a txn but it has not executed any SQL for 5 min (that is, it never executes `COMMIT`), will the backend be recycled? 
If it will be, then what if it continues the transaction after the backend is recycled?

pkg/manager/router/backend_observer.go
```diff
@@ -160,6 +162,7 @@ func (bo *BackendObserver) observe(ctx context.Context) {
 			select {
 			case <-time.After(wait):
 			case <-bo.refreshChan:
+				refresh = true
```
djshow832, line 165: Gateway may call `Router.RefreshBackend` to refresh. I don't remember when Gateway may call it. Please ensure that in that case, adding one backend back is expected.
Besides, the word `refresh` is vague. Every time I'll think for a while what the parameter means.

pkg/manager/router/router_score.go
clark1013: Nope, If a connection is in a transaction the backend will not quit until the transaction be commit. Controlled by inTransCount in this line. https://github.com/tidbcloud/tidb-cse/blob/release-7.1-keyspace/standby/idle_watcher.go#L83

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -337,6 +348,13 @@ func (mgr *BackendConnManager) ExecuteCmd(ctx context.Context, request []byte) (
 	if mgr.closeStatus.Load() >= statusClosing {
 		return
 	}
+	if mgr.backendIO.Load() == nil {
+		err = mgr.resume(ctx)
+		if err != nil {
+			mgr.logger.Info("resume failed", zap.Error(err))
```
djshow832, line 354:
```suggestion
			mgr.logger.Error("resume failed", zap.Error(err))
```

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -509,20 +529,25 @@ func (mgr *BackendConnManager) tryRedirect(ctx context.Context) {
...
-		// If the backend connection is closed, also close the client connection.
-		// Otherwise, if the client is idle, the mgr will keep retrying.
-		if errors.Is(rs.err, net.ErrClosed) || pnet.IsDisconnectError(rs.err) || errors.Is(rs.err, os.ErrDeadlineExceeded) {
-			mgr.quitSource = SrcBackendNetwork
-			if ignoredErr := mgr.clientIO.GracefulClose(); ignoredErr != nil {
-				mgr.logger.Warn("graceful close client IO error", zap.Stringer("client_addr", mgr.clientIO.RemoteAddr()), zap.Error(ignoredErr))
+	if backendIO == nil {
+		mgr.logger.Info("session paused, skip redirect")
+		return
+	} else {
```
djshow832, line 535: You have returned if `backendIO == nil`, so no need to add `else`.

pkg/manager/router/backend_observer.go
clark1013: I have stopped Gateway to call RefreshBackend in my PR and it's works well. The router call GetBackendList every 3 seconds. It's fast enough for Gateway.
How about "force"? it sounds more natural in serverlerss scenario. But I am not sure if it make sense in TiProxy.

pkg/manager/router/router_score.go
djshow832: OK, then it'll be fine. Leave a comment here to clarify that when a connection is redirecting, the backend won't be paused.

pkg/manager/router/backend_observer.go
djshow832: No, just add a comment on the function definition.

pkg/manager/router/router_score.go
```diff
@@ -318,6 +395,36 @@ func (router *ScoreBasedRouter) OnBackendChanged(backends map[string]*BackendHea
...
+					switch conn.phase {
+					case phasePauseNotify:
+						continue
+					case phaseRedirectNotify:
+						router.logger.Info("connection is redirecting, maybe a new backend is alive. skip pause")
+						backend.connScore++
+						continue
+					}
+					conn.phase = phasePauseNotify
+					conn.Pause()
```
djshow832, line 423:
```suggestion
					conn := ele.Value
					switch conn.phase {
					case phasePauseNotify:
						continue
					case phaseRedirectNotify:
						router.logger.Info("connection is redirecting, maybe a new backend is alive. skip pause")
						continue
					}
					backend.connScore--
					conn.phase = phasePauseNotify
					conn.Pause()
```

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +828,98 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+func (mgr *BackendConnManager) Pause() bool {
+	if mgr.closeStatus.Load() >= statusNotifyClose {
+		return false
+	}
+
+	if mgr.backendIO.Load() == nil {
+		return false
+	}
+
+	mgr.signalReceived <- signalTypePause
```
djshow832, line 842: It may block here if the previous signal is not processed yet. Thus, the whole pause procedure in the `Router` may block.

clark1013: If `Router` blocks here, `BackendObserver` can be blocked too. Then backend list can not be flushed correctly. Maybe it's better to do perform the pause procedure in another goroutine. Let me take a look carefully.

djshow832: Oh, I was wrong. It won't block here because the channel size is big enough and the same signal won't be sent again before it's processed.
Never mind.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -640,6 +669,9 @@ func (mgr *BackendConnManager) checkBackendActive() {
 		return
 	}
 	backendIO := mgr.backendIO.Load()
+	if backendIO == nil && mgr.lastPauseTime != 0 && monotime.Since(mgr.lastPauseTime) < mgr.config.PausedConnTimeout {
```
djshow832: Why use `&&`?
If `backendIO==nil` and the other conditions are unsatisfied, the next line will panic.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +828,98 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+	if err != nil {
+		mgr.logger.Error("query session failed when pause", zap.Error(err))
+		mgr.notifyPauseResult(mgr.ServerAddr(), err)
+		return
+	}
+	mgr.sessionState.Store(&SessionState{
+		sessionStates: sessionStates,
+		sessionToken:  sessionToken,
+	})
+	mgr.backendIO.Store(nil)
```
djshow832, line 873: The backendIO is not closed.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +828,98 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+// notify pause result to eventReceiver.
+func (mgr *BackendConnManager) notifyPauseResult(addr string, pauseErr error) {
+	eventReceiver := mgr.getEventReceiver()
+	if eventReceiver == nil {
+		return
+	}
+	if pauseErr != nil {
+		err := eventReceiver.OnPauseFail(addr, mgr)
+		mgr.logger.Warn("pause connection failed", zap.String("addr", addr),
+			zap.NamedError("redirect_err", pauseErr), zap.NamedError("notify_err", err))
```
djshow832, line 887:
```suggestion
			zap.NamedError("pause_err", pauseErr), zap.NamedError("notify_err", err))
```

pkg/proxy/backend/backend_conn_mgr.go
clark1013: Yes, I have found it during test. It has been fixed.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +835,101 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+		err := errors.New("session state is nil")
+		return err
+	}
+	err := json.Unmarshal([]byte(state.sessionToken), token)
+	if err != nil {
+		return err
+	}
+
+	clientResp := pnet.HandshakeResp{User: token.Username}
+	mgr.logger.Info("start to resume session")
```
djshow832, line 912: Too many logs in this function.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +835,101 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+		mgr.handshakeHandler.OnHandshake(mgr, newBackendIO.RemoteAddr().String(), err, Error2Source(err))
+	}
+	mgr.logger.Debug("finish to handshakeSecondTime")
+	if err != nil {
+		if ignoredErr := newBackendIO.Close(); ignoredErr != nil && !pnet.IsDisconnectError(ignoredErr) {
+			mgr.logger.Error("close new backend connection failed", zap.Error(ignoredErr))
+		}
+		return err
+	}
+	mgr.backendIO.Store(newBackendIO)
```
djshow832, line 930: `mgr.backendIO` is already set in `getBackendIO`, so you should set it to nil when err != nil.

pkg/proxy/backend/backend_conn_mgr.go
```diff
@@ -798,3 +835,101 @@ func (mgr *BackendConnManager) ConnInfo() []zap.Field {
...
+	if mgr.backendIO.Load() == nil {
+		return false
+	}
+
+	mgr.signalReceived <- signalTypePause
+	return true
+}
+
+// tryPause read sessionStates from backend and store the states.
+func (mgr *BackendConnManager) tryPause(ctx context.Context) {
```
djshow832, line 854: `updateAuthInfoFromSessionStates` is not called in either pause or resume.

pkg/proxy/backend/backend_conn_mgr.go
djshow832: Call `mgr.handshakeHandler.OnHandshake(mgr, mgr.ServerAddr(), nil, SrcNone)` after success.

pkg/proxy/backend/backend_conn_mgr.go
clark1013: I removed the last two one, it's enough to log resume starting as key point.

pkg/proxy/backend/backend_conn_mgr.go
clark1013: Fixed. I also called updateTraffic as Redirect do. PTAL if that's corret.

djshow832 (approved):

Issue #642, opened 2024-08-26
Support traffic replay


djshow832:
### Describe your feature request related problem

A new TiDB version may have some compatibility breakers, such as:
- A statement runs may unexpectedly report an error on the new version
- A statement may result in different query results on the new version
- A statement may have a worse plan and runs slower on the new version

Another scenario is that when the cluster runs unexpectedly, such as very slowly, the users may want to capture the traffic so that they can investigate by replaying the traffic after recovering the database.

There exist some database traffic replay tools, including tcpcopy, mysql replay, and query-playback. Tcpcopy and mysql replay capture data by the way similar to tcpdump, while query-playback is based on slow log. Although some of them are built for MySQL, it's fine to deploy them on the proxy instance to capture and replay the overall traffic.

However, they have some limitations:
- Users need to learn the deployment and usage of these tools.
- Tcpcopy only captures new connections, which is not friendly for preserving connections. Mysql replay can capture existing connections but it loses session states such as the current schema and prepared statements, which may make it not work.
- Tcpcopy and mysql replay capture the encrypted data when TLS is enabled. The encrypted data can not be replayed on another cluster.
- Users need to verify the results and performance by themselves.

### Describe the feature you'd like

Capturing traffic on one TiDB cluster and replaying the traffic on a new TiDB cluster to verify the SQL compatibility and performance of the new TiDB cluster.

### Describe alternatives you've considered

None

### Teachability, Documentation, Adoption, Migration Strategy

None

PR #643, opened 2024-08-27
RFC: Traffic Capture and Replay

djshow832:

Issue Number: ref #642 

Problem Summary:
There are some cases when users want to capture the traffic on the production cluster and replay the traffic on a testing cluster:
- A new TiDB version may have compatibility breakers, such as the statements failing, running slower, or resulting in different query results.
- When the cluster runs unexpectedly, users want to capture the traffic so that they can investigate it later by replaying the traffic.
- Test the maximum throughput of a scaled-up or scaled-down cluster using the real production workload instead of standard bench tools.

This proposes a design of capturing traffic on the production cluster and replaying the traffic on a testing cluster to verify the SQL compatibility and performance of the new cluster.

### Check List

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


docs/design/2024-08-27-traffic-replay.md
```diff
@@ -0,0 +1,231 @@
...
+# Cmd_type: 3
+# Rows_sent: 1
+# Result_size: 100
+# Payload_len: 8
+SELECT 1
+```
+
+To replay the statements in the same order, the statements should be ordered by the start time. So the field `Time` indicates the start time instead of the end time. To order by the start time, one statement must stay in memory until all the previous statements are finished. If a statement runs too long, it's logged without query result information, so the replay skips checking the result and duration.
+
+It's trivial to lose traffic during failure, so TiProxy doesn't need to flush the traffic synchronously. To reduce the effect on the production cluster, TiProxy flushes the traffic in batch asynchronously with double buffering. Similarly, TiProxy loads traffic with double buffering during replaying. If the IO is too slow and the buffer accumulates too much, the capture should stop immediately and the status should be reported in UI and Grafana.
```
bb7133, line 83: Could we estimate the performance impact of captures/replays to TiProxy?

djshow832: Yes, latency regression < 1% and throughput < 5%. But I think it should be specified in the spec, not the design.

xhebox (approved):

PR #949, opened 2025-10-20
cmd, replay: allocate connection id in local decoder

### LINKED ISSUE #948: The connection count may drop after creating 1M connections in sql replay
## Bug Report

Please answer these questions before submitting your issue. Thanks!

### 1. Minimal reproduce step (Required)

The upstream cluster enables 32-bits connection ID.
Replay audit logs and wait until creating 1M connections.

### 2. What did you expect to see? (Required)

Connection count is right

### 3. What did you see instead (Required)

Connection count drops

### 4. What is your version? (Required)

master

YangKeao:

Issue Number: close #948

Problem Summary:

1. If a connection closed, the following commands with the same connection id will not be executed.

1. Modify the `decoder` to generate a new unique connection id for each new connection.
2. Modify the replay connection to wait for `ComQuit` to finally close the goroutine. Before that, all commands except `ComQuit` will be ignored.

It costs more than previous situation: the goroutine will not be freed until `ComQuit` comes. However, I don't think it's big thing.
Another possible solution is to stop the goroutine and handle the status carefully outside, but I think it's more difficult to handle the lock and more difficult to test, so I chose this solution.

If `ComQuit` never appears, the goroutine will leak, which is just the same for non-fault cases.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```


pkg/sqlreplay/conn/conn.go
```diff
@@ -136,6 +142,13 @@ func (c *conn) Run(ctx context.Context) {
 			if command == nil {
 				break
 			}
+			if c.waitForQuitMode {
+				if command.Value.Type == pnet.ComQuit {
+					return
+				}
+
+				continue
```
djshow832: Add the count of c.replayStats.ExceptionCmds.

djshow832 (commented):
Rest LGTM

PR #972, opened 2025-11-06
sqlreplay, backend: reconnect after a replaying connection disconnects

### LINKED ISSUE #964: Init DB when replay instead of parsing
## Development Task

There are 2 problems when generating InitDB when parsing:
- The CPS may differ from the production cluster because the production cluster generally inits DB in handshake
- When a connection disconnects, the subsequent commands in the connection are discarded, resulting in QPS drop during TiDB restarting.

So the better way is to initializing DB in the handshake when the backend connection creates or reconnects.

djshow832:

Issue Number: close #964 

Problem Summary:
There are 2 problems when generating InitDB when parsing:
- The CPS may differ from the production cluster because the production cluster generally inits DB in handshake
- When a connection disconnects, the subsequent commands in the connection are discarded, resulting in QPS drop during TiDB restarting.

So the better way is to initializing DB in the handshake when the backend connection creates or reconnects.

- Do not generate InitDB command in the decode phase, set the CurDB in the command instead
- In the replay phase, handshake with the initial DB or generate InitDB command if necessary
- Remove the waitForQuitMode, reconnect before the next command if the connection disconnects
- If a prepared statement is not found, do not send it to the server, just generate an exception

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

1. replay
2. kill tidb and start another tidb
3. check the failure count

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/proxy/backend/authenticator.go
+++ b/pkg/proxy/backend/authenticator.go
@@ -301,7 +301,7 @@ func (auth *Authenticator) handshakeWithBackend(ctx context.Context, logger *zap
 	}
 	auth.user = username
 	auth.dbname = dbName
-	auth.capability = handshakeHandler.GetCapability()
+	auth.capability = handshakeHandler.GetCapability() | pnet.ClientConnectWithDB
 	auth.collation = pnet.Collation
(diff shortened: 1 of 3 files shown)


YangKeao (approved):
LGTM

PR #1141, opened 2026-04-21
api: add `PUT/DELETE /health` endpoint to mark the instance as unhealthy

### LINKED ISSUE #1140: Add an API to mark a TiProxy instance as unhealthy
## Development Task

To support gracefully shutdown a TiProxy, we need to add an API to mark a TiProxy instance as unhealthy so that the LB (usually from cloud provider) will avoid assigning new connections to it while keeping the existing connections.

YangKeao:

Issue Number: close #1140

Problem Summary:

For the LB from cloud provider, once the Pod enters `terminating` status, the connection will persist at most 15min (for aliyun) and 1h (for AWS). However, for `unhealthy`, the connection can be persisted for a very long time: 100h for AWS and not limited for Aliyun (I've only tested for 1h, and I'll try 24h tests later).

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/server/api/debug.go
+++ b/pkg/server/api/debug.go
@@ -13,12 +13,25 @@ import (
 
 func (h *Server) DebugHealth(c *gin.Context) {
 	status := http.StatusOK
-	if h.isClosing.Load() || !h.mgr.NsMgr.Ready() {
+	health := config.HealthInfo{
+		ConfigChecksum: h.mgr.CfgMgr.GetConfigChecksum(),
+	}
+	if h.unhealthyMark.Load() {
+		status = http.StatusBadGateway
+	} else if h.isClosing.Load() || !h.mgr.NsMgr.Ready() {
 		status = http.StatusBadGateway
 	}
-	c.JSON(status, config.HealthInfo{
-		ConfigChecksum: h.mgr.CfgMgr.GetConfigChecksum(),
-	})
+	c.JSON(status, health)
+}
+
+func (h *Server) DebugSetHealthUnhealthy(c *gin.Context) {
+	h.unhealthyMark.Store(true)
+	c.JSON(http.StatusOK, "")
+}
+
+func (h *Server) DebugUnsetHealthUnhealthy(c *gin.Context) {
+	h.unhealthyMark.Store(false)
+	c.JSON(http.StatusOK, "")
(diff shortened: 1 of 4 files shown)


pkg/server/api/debug.go
```diff
@@ -5,20 +5,56 @@ package api
...
+		status = http.StatusBadGateway
+		health.UnhealthyReason = "server is closing"
+	} else if !h.mgr.NsMgr.Ready() {
 		status = http.StatusBadGateway
+		health.UnhealthyReason = "namespace manager is not ready"
 	}
-	c.JSON(status, config.HealthInfo{
-		ConfigChecksum: h.mgr.CfgMgr.GetConfigChecksum(),
-	})
+	c.JSON(status, health)
```
djshow832, line 37: If it's set healthy manually, the status is still OK even when TiProxy is closing, is that expected?

pkg/server/api/debug.go
```diff
@@ -5,20 +5,56 @@ package api
...
+		if !healthOverride.Healthy {
+			status = http.StatusBadGateway
+			health.UnhealthyReason = healthOverride.Reason
+		}
+	} else if h.isClosing.Load() {
+		status = http.StatusBadGateway
+		health.UnhealthyReason = "server is closing"
+	} else if !h.mgr.NsMgr.Ready() {
 		status = http.StatusBadGateway
+		health.UnhealthyReason = "namespace manager is not ready"
```
djshow832:
```suggestion
		health.UnhealthyReason = "server is not ready"
```

pkg/server/api/debug.go
YangKeao: Yes. I think it's more intuitive to allow this interface to overwrite the status in any direction: from `healthy` to `unhealthy` and from `unhealthy` to `healthy` (and this behavior is also covered by tests).

Though for now, we don't have any scenario that we need to use this interface to set it to healthy manually :thinking: (if we want to resume from manually set `unhealthy`, we can use `DELETE` instead of setting `healthy`).

djshow832 (approved):

PR #1185, opened 2026-06-23
balance: read metrics directly from backends when failing to query from owners

### LINKED ISSUE #768: Fallback to connection-based balance when any tidb misses CPU usage
## Development Task

I've encountered an issue where one TiDB node doesn't have CPU or memory limit. The CPU and memory usage of this node are assumed to be 100%, and no connections will be routed to it.

This may cause problems, so it's better to fallback to connection-based balance in this situation. Besides, warnings should be logged for troubleshooting.

djshow832:

Issue Number: close #768 

Problem Summary:
Some problems may cause TiProxy to fail to query some backend metrics:
- The owner is not the right owner
- Network between TiProxy and the owner
In such cases, the backend CPU usage may be imbalance.

If some backend metrics are missing, read directly from the backends.

### Check List

Tests 

- [x] Unit test
- [ ] Integration test
- [ ] No code

Notable changes

- [ ] Has tiproxyctl change

### Release note

```release-note
None
```

### DIFF
--- a/pkg/balance/metricsreader/backend_reader.go
+++ b/pkg/balance/metricsreader/backend_reader.go
@@ -216,6 +216,7 @@ func (br *BackendReader) ReadMetrics(ctx context.Context) error {
 		}
 	}
 
+	// If the metrics of some backends are missing, read directly from the backends.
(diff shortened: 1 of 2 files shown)


YangKeao (approved):
LGTM

PR #1204, opened 2026-08-18
sqlreplay: retry transient traffic file listing failures

YangKeao:
## Problem

In dynamic replay with `WaitOnEOF`, a transient S3 `ListObjects` timeout caused the per-directory rotate reader to exit and close its file channel. The merge decoder then observed EOF and permanently removed that input while the directory watcher still considered the directory registered.

This made replay QPS fall as more readers timed out. The previous 10ms empty-tail polling also generated excessive S3 list requests.

## Changes

- Retry file listing failures at a 1s interval only when `WaitOnEOF` is enabled.
- Keep the existing 10s deadline for each file listing attempt.
- Change empty-tail polling from an unconditional 10ms sleep to a context-aware 1s wait.
- Preserve the existing behavior for file open failures and non-live replay.
- Add one regression test that injects `context deadline exceeded` and verifies that the same reader recovers and reads the next file.

## Tests

- `go test -p=1 ./pkg/sqlreplay/store -count=1`
- `go test -p=1 -race ./pkg/sqlreplay/store -run TestWaitOnEOFRecoversFromWalkError -count=1`
- `go test -p=1 ./pkg/sqlreplay/replay -run 'TestDynamicInput|TestMergeDecoder' -count=1`\n\nFull-repository lint was not run due to the available memory limit.

### DIFF
--- a/pkg/sqlreplay/store/const.go
+++ b/pkg/sqlreplay/store/const.go
@@ -14,4 +14,7 @@ const (
 	fileSize            = 300 << 20
 	bufferSize          = 1 << 20
 	opTimeout           = 10 * time.Second
+	readerPollInterval  = time.Second
+	readerRetryInitial  = time.Second
+	readerRetryMax      = 30 * time.Second
 )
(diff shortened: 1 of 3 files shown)


YangKeao:
Not ready yet. Close.

YangKeao:
Superseded by #1205. The replacement PR contains the narrowed, single-purpose change (100 additions / 3 deletions); this PR remained pinned to the pre-narrowing head SHA after the fork branch was force-pushed.

PR #1205, opened 2026-08-18
sqlreplay: retry transient traffic storage failures

YangKeao:
## Problem

A transient S3 `ListObjects` or file `Open` failure caused the rotate reader to exit and close its file channel. The upper layer then observed EOF instead of the storage error and permanently removed that input. This could silently reduce replay QPS or truncate a non-live replay.

## Changes

- Use one shared retry path for file listing and file open failures in both `WaitOnEOF` modes.
- Return EOF with `WaitOnEOF=false` only after a successful listing confirms that no next file exists.
- Keep retrying the selected file after an `Open` failure instead of listing again, so a later file cannot be replayed first.
- Keep the existing 10s deadline for each file listing attempt.
- Keep the original 10ms polling behavior after a successful empty listing.
- Log the sanitized storage URI during retries instead of the raw configured URI, which may contain credentials.
- Add regression tests for transient listing and open failures, including `WaitOnEOF=false`.

## Tests

- `go test -p=1 ./pkg/sqlreplay/store -count=1`
- `go test -p=1 -race ./pkg/sqlreplay/store -run "TestWaitOnEOF|TestRecoversFromWalkError|TestWaitOnEOFRetriesSameFileAfterOpenError|TestWalkS3" -count=3`
- `go test -p=1 ./pkg/sqlreplay/replay -run "^(TestDynamicInput|TestDynamicInputCoverAllDirectories|TestMergeDecoderWaitingDecoder|TestMergeDecoderDecodeError|TestMergeDecoderCancel)$" -count=3`
- MinIO recovery check: paused MinIO long enough to trigger two consecutive 10s `ListObjects` deadlines, resumed it, verified a `WaitOnEOF=false` reader consumed the expected object and returned EOF only after a successful empty listing. The retry logs contained only the sanitized bucket/prefix URI.

A broader `go test -p=1 ./pkg/sqlreplay/... -count=1` run was attempted but was not clean because of unrelated existing failures in `TestLogAndSkipInvalidExecuteRequest` and the randomized merge-decoder test; the long-running conn package was stopped after more than four minutes.

Full-repository lint was not run due to the available memory limit.

### DIFF
--- a/pkg/sqlreplay/store/const.go
+++ b/pkg/sqlreplay/store/const.go
@@ -14,4 +14,5 @@ const (
 	fileSize            = 300 << 20
 	bufferSize          = 1 << 20
 	opTimeout           = 10 * time.Second
+	readerRetryInterval = time.Second
 )
--- a/pkg/sqlreplay/store/rotate.go
+++ b/pkg/sqlreplay/store/rotate.go
@@ -296,6 +296,9 @@ func (r *rotateReader) openFileLoop(ctx context.Context) error {
 			})
 		cancel()
 		if err != nil {
+			if r.waitForStorageRetry(ctx, "list traffic files", err) {
+				continue
+			}
 			break
 		}
 		if minFileName == "" {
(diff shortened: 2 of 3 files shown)


bb7133 (approved):
LGTM

djshow832 (approved):

PR #1226, opened 2026-09-15
proxy: return an error on COM_PING during graceful shutdown

### LINKED ISSUE #1224: Change COM_PING response during graceful shutdown
### Describe your feature request related problem

For applications like [keycloak](https://github.com/keycloak/keycloak) it is difficult to efficiently and correctly detect what connections should no longer be used when TiProxy is used.

TiDB returns an error result for `COM_PING` when it is doing a graceful shutdown.

### Describe the feature you'd like

Change the `COM_PING` response during graceful shutdown to return an error.

### Teachability, Documentation, Adoption, Migration Strategy

Related: https://github.com/keycloak/keycloak/issues/49558#issuecomment-5607571444

dveeden:

Issue Number: close #1224

Problem Summary:

Applications and connection pools (e.g. [Keycloak](https://github.com/keycloak/keycloak/issues/49558#issuecomment-5607571444)) use `COM_PING` to decide whether a pooled connection is still usable. TiDB replies to `COM_PING` with an error once it starts a graceful shutdown, so clients can drain their pools. TiProxy forwarded the ping to the backend and returned OK, so clients kept using connections to an instance that was about to go away.

TiProxy now answers `COM_PING` the same way TiDB does. For reference, `pkg/server/conn.go` in TiDB:

```go
case mysql.ComPing:
	if cc.server.health.Load() {
		return cc.writeOK(ctx)
	}
	return servererr.ErrServerShutdown
```

- `SQLServer` gets a `shuttingDown` flag, set at the very beginning of `PreClose()`. TiDB sets its `health` flag before sleeping for `graceful-wait-before-shutdown`, so TiProxy does the same — the ping must already report the drain during that window, which is exactly when the load balancer is taking the instance offline.
- The flag is passed to each connection through the new `BCConfig.ShuttingDown` closure, and `BackendConnManager.ExecuteCmd()` answers `COM_PING` itself instead of forwarding it to the backend.
- The reply is `ER_SERVER_SHUTDOWN` (1053, SQLSTATE `08S01`, "Server shutdown in progress"), identical to TiDB. As in TiDB, the connection is **not** closed: a MySQL error is returned to the client and the connection keeps serving until the existing graceful-close logic tears it down.

`ShuttingDown` is nil for the traffic replay path, so replayed pings are unaffected.

### Check List

Tests

- [x] Unit test
- [ ] Integration test
- [ ] No code

`TestPingDuringShutdown` checks that the ping is forwarded to the backend while serving, and that once shutting down TiProxy replies with 1053/`08S01` without sending anything to the backend, keeps the connection alive, and still forwards other commands.

`TestShuttingDownBeforeGracefulWait` checks that the flag is set at the start of `PreClose`, not after `graceful-wait-before-shutdown`.

Notable changes

- [ ] Has tiproxyctl change
- [x] Other user behavior changes

### Release note

```release-note
TiProxy now returns an error for `COM_PING` during graceful shutdown, the same as TiDB, so that clients and connection pools can detect that the instance is draining.
```

## Summary by CodeRabbit

* **New Features**
  * During graceful shutdown, ping requests now receive a server-shutdown error instead of being forwarded to the backend.
  * Existing connections remain open while reporting the shutdown state, allowing clients to handle it gracefully.

* **Tests**
  * Added coverage for ping behavior during shutdown and verification that shutdown status is reported before the server fully closes.


dveeden:
Note: This needs testing

djshow832 (approved):

